PR #42, opened 2018-02-13
FIX: docs and sub-Devices in DeviceDisplay

klauer:
Noted these issues while poking around on a local installation. Worked quite well otherwise!

I'll send along more details on my little example later, as I think you'll find it interesting.


teddyrendahl:
🏆 The first PR from someone who wasn't me!

No brainer on each point. Thanks!

PR #45, opened 2018-02-27
CI: Installation via Conda

teddyrendahl:
## Description

When I tagged Typhon `v0.1.0` we did not have a dedicated Conda channel for the PCDS group so I didn't bother making a recipe. We now have `pcds-tag` and `pcds-dev` so this PR builds Typhon and deploys to the appropriate channel.

Also since creating the repository @ZLLentz laid out a way to do the Travis installation entirely via Conda. I think this is a more useful testing strategy as:
1) Most users will install via the Conda route
2) In prior iterations we would test and evaluate pull requests based on the `pip` install then only when the PR was merged would we see that the conda-recipe was invalid and the build would fail.

Finally, I added a build to the matrix that tests with `pydm-dev` instead of the tagged builds. This will allow us to use new features that haven't yet been tagged by the `PyDM` team, as well as give them some useful feedback if they break something unexpected.

### Side Note
 I kind of want to host a PyDM build on the `pcds-*` channels. It is slightly annoying to have a Conda channel to just include a single package. We as a group will also want to host a `pyepics` install so the nice thing is it would then avoid `gsecars`...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -92,7 +92,7 @@ after_success:
       anaconda upload bld-dir/linux-64/*.tar.bz2
     fi
   - |
-    if [[ $TRAVIS_PULL_REQUEST == false && $TRAVIS_REPO_SLUG == $OFFICIAL_REPO && $TRAVIS_BRANCH == 'master' && $TRAVIS_TAG == '' && $BUILD ]]; then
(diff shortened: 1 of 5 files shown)


teddyrendahl:
Slight change breaking with our normal convention. I want to build the `-dev` build even during tags. This removes the situation where the `-tag` build is actually ahead of `-dev`

PR #47, opened 2018-03-04
ENH: SignalPlugin

### LINKED ISSUE #44: SignalPlugin
## Expected Behavior

A plugin that uses Ophyd's subscription system to connect to the standard PyDMChannel slots. The channel will be specified with `sig://{signal_type}/{args}/{kwargs}`.

## Current Behavior

Signals are examined and then the pvnames are ripped out of them and stuffed in channel access.

## Context

The current setup precludes the ability to display any other kind of signal. It also depends on the API of which ever Ophyd `control_layer` you choose to use. We should keep the abstraction at the Ophyd layer and not make assumptions about the control_layer underneath.

teddyrendahl:
## Description

Included are two custom `Plugin` types.
* `ClassPlugin` a generic way to load arbitrary objects with an address. This won't be called directly but stores the common initialization that is done by `SignalPlugin` and will eventually be done by `TyphonPlugin` and `HappiPlugin`.
* `SignalPlugin` subscribes to any Ophyd Signal and sends updates to the listening channels. It also allows writes back to the signal itself.

I don't imagine either of these plugins will be used in screens themselves. I therefore didn't make the actual `PyDMConnection` objects. The shenanigans of correctly typing `args` and `kwargs` from the address doesn't seem like a road we should go down. `TyphonPlugin` will just inspect the Ophyd object and create the signals using the `from_object` classmethod.

### Outstanding Questions
* There is not a good way to tell if a `Signal` can be written to. We are forced to `try / except`
* We can expect all `Signal` objects to have a `connected` attribute. Should we send this information back through the channel?
* `Signal.set` or `Signal.put`?

## Motivation and Context

Closes #44 

## How Has This Been Tested?

* Instantiation of a `ClassConnection` object
* Connect and disconnect a `SignalPlugin`

Short table of Plugin information

### DIFF
--- a/typhon/plugins/core.py
+++ b/typhon/plugins/core.py
@@ -19,6 +19,52 @@ from pydm.PyQt.QtCore import pyqtSlot, Qt
 logger = logging.getLogger(__name__)
 
 
+def obj_from_string(clsname, args=None, kwargs=None):
+    """
+    Create an object from a string specification
+
+    Provide a name of a class as well as optional arguments and keywords. This
+    class handles importing the class and instantiating a new object.
+
+    Parameters
+    ----------
+    clsname : str
+
+    args: optional
+        Passed directly to class constructor
+
+    kwargs: optional
+        Passed directly to class constructor
+
+    Returns
+    -------
+    obj: object
+        Object of type ``clsname``
+    """
+    # Import the class
+    mod, cls = clsname.rsplit('.', 1)
+    if mod in sys.modules:
+        mod = sys.modules[mod]
+    else:
+        logger.debug("Importing %s", mod)
+        mod = importlib.import_module(mod)
+    # Grab our device
+    cls = getattr(mod, cls)
+    # Format arguments
+    if args:
+        args = args.split(',')
+    else:
+        args = list()
+    # Format keywords
+    if kwargs:
+        kwargs = dict([pair.split('=', 1)
+                       for pair in kwargs.split(',')])
+    else:
+        kwargs = dict()
+    # Create object
+    return cls(*args, **kwargs)
+
+
 class ClassConnection(PyDMConnection):
     """
     Connection which spawns an object of specified class
@@ -53,8 +99,6 @@ class ClassConnection(PyDMConnection):
             logger.debug("Using an already instantiated object as reference")
             self.obj = preassembled
         else:
-            # NOTE: This was taken entirely from happi/loader.py. The reason it
-            # was not directly imported was to avoid an unneeded dependency
             # Parse the classname, arguments and keywords from the address
             # First assume the form {class}|{args}|{kwargs}
             try:
@@ -63,29 +107,9 @@ class ClassConnection(PyDMConnection):
             except ValueError:
                 cls, args = address.split('|', 1)
                 kwargs = None
-            # Import the class
-            mod, cls = cls.rsplit('.', 1)
-            if mod in sys.modules:
-                mod = sys.modules[mod]
-            else:
-                logger.debug("Importing %s", mod)
-                mod = importlib.import_module(mod)
-            # Grab our device
-            cls = getattr(mod, cls)
-            # Format arguments
-            if args:
-                args = args.split(',')
-            else:
-                args = list()
-            # Format keywords
-            if kwargs:
-                kwargs = dict([pair.split('=', 1)
-                               for pair in kwargs.split(',')])
-            else:
-                kwargs = dict()
             # Create our object
-            logger.debug("Instantiating %s ...", cls.__name__)
-            self.obj = cls(*args, **kwargs)
+            logger.debug("Instantiating %s ...", cls)
+            self.obj = obj_from_string(cls, args, kwargs)
 
     @classmethod
     def from_object(cls, obj):
--- a/typhon/plugins/core.py
+++ b/typhon/plugins/core.py
@@ -32,8 +32,8 @@ class ClassConnection(PyDMConnection):
     -------
     .. code:: python
 
-        conn = ClassConnection(cls://ophyd.EpicsMotor|'Tst:Mtr:07'|name=Test,
-                               ophyd.EpicsMotor|'Tst:Mtr:07'|name=Test)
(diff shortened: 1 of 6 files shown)


typhon/plugins/core.py
```diff
@@ -0,0 +1,198 @@
...
+    ----------
+    obj: object
+        The created object is stored on the connection for use by subclasses
+
+    Example
+    -------
+    .. code:: python
+
+        conn = ClassConnection(cls://ophyd.EpicsMotor|'Tst:Mtr:07'|name=Test,
+                               ophyd.EpicsMotor|'Tst:Mtr:07'|name=Test)
```
hhslepicka: Shouldn't it be this instead?
```
conn = ClassConnection("cls://ophyd.EpicsMotor|'Tst:Mtr:07'|name=Test",
                       "ophyd.EpicsMotor|'Tst:Mtr:07'|name=Test")
```

teddyrendahl:
Thanks for the review Hugo!

1. I'm not sure what to do about this. If you are just giving me objects I don't think I can cache them very easily. This will be solved by the `happi` plugin where caching is already implemented.
2. This means just implementing the `send_connection_state` , `send_access_state`? This kind of leads into one of my questions. How do I tell if a signal is writable without `try / except`?

hhslepicka:
1. If the cache is already implemented by `happi` you are fine. This case is the same as `pyepics` in which the PV object is cached (properly or not is a different story 🤣 )

2. Yes, you need to send both. Unfortunately some cases are like that and you just simply give the default of: Yes, you can write and if you get an exception you make it read-only... not the best design ever but that all happens by the fact that the signal has no info about it. In your case you can check if it is a SignalRO or Signal... if it is an `EpicsSignal` this information can be added at the `Ophyd` level by making use of the `access callback`.

typhon/plugins/core.py
```diff
@@ -0,0 +1,198 @@
...
+        # Base initialization
+        super().__init__(channel, address, protocol=protocol, parent=parent)
+        # Just use the preassambled object if given to us
+        # Otherwise, instantiate our own
+        logger.debug("Creating connection to %s", address)
+        if preassembled:
+            logger.debug("Using an already instantiated object as reference")
+            self.obj = preassembled
+        else:
+            # NOTE: This was taken entirely from happi/loader.py. The reason it
```
klauer: Could this then be brought out to a supporting function? The parsing probably should not be buried in this `__init__` (even if it is copied from `happi`) and tested separately.

If these functions are ever moved up into a shared package, maybe then a grammar to parse would be a good idea (current one is problematic with no way to escape`|`, commas/equals in strings, etc...)

typhon/plugins/core.py
```diff
@@ -0,0 +1,198 @@
...
+    Attributes
+    ----------
+    obj : ophyd.Signal
+        Stored signal object
+
+    Example
+    -------
+    .. code:: python
+
+        conn = ClassConnection(sig://ophyd.Signal|name=Test,)
```
klauer: Missing quotes here

typhon/plugins/core.py
```diff
@@ -0,0 +1,198 @@
...
+        """
+        Pass a value from the UI to Signal
+
+        We are not guaranteed that this signal is writeable so catch exceptions
+        if they are created
+        """
+        try:
+            self.obj.put(new_val)
+        except Exception as exc:
+            logger.exception("Unable to put %s to %s", new_val, self.obj.name)
```
klauer: Consider `%r` for `new_val`

@hhslepicka is logging in PyDM moving along yet? Would be great to have this info propagated back to the user outside of just the terminal.

typhon/plugins/core.py
```diff
@@ -0,0 +1,198 @@
...
+        This removes the `send_new_value` connections from the channel in
+        addition to the default disconnection performed in PyDMConnection
+        """
+        # Disconnect put_value from outgoing channel
+        if channel.value_signal is not None:
+            for _typ in self.supported_types:
+                try:
+                    channel.value_signal[_typ].disconnect(self.put_value)
+                except KeyError:
+                    logger.debug("Unable to disconnect value_signal from %s "
```
klauer: Missing `%s` here

typhon/plugins/core.py
hhslepicka: @klauer PyDM at master sets up the logger, we still need to make the usage of logging the default at all code instead of prints... There is still work to do but it works.

teddyrendahl: Thanks @klauer! I know that this will use `repr` instead of `str` but what do you expect from `str` that you think will not be ideal? Just curious as I usually both interchangeably.

typhon/plugins/core.py
teddyrendahl: @klauer Thanks for calling me out on this. I think I'll just add a dependency on `happi` where this function is tested thoroughly. If someone really wants to use `typhon` but not incur the innocuous `happi` dependency we'll cross that bridge then.

teddyrendahl: I looked at this again, and there are actually some deep seeded differences that I forgot about. I refactored the object instantiation out of the class constructor and into a new function and added a test.

teddyrendahl:
This should be ready for follow up:

1) Add `Plugin` objects but didn't set them in up in PyDM until https://github.com/slaclab/pydm/issues/307 is resolved
2) `ClassConnection` automatically reports objects as accessible and connected because there isn't a way to determine this from a generic object. `SignalConnection` was more difficult and I wasn't certain what to do. The generic `Signal` object doesn't have a connection callback but there is a `connected` state.

typhon/plugins/core.py
klauer: Sorry for the late reply - I like `repr` here because then the exception message makes sense regardless of the value `new_val` contained. Strings, for example, get wrapped in quotes with `repr` and not with `str`. For user types, it assumes your classes have set `__repr__` and `__str__` appropriately, of course...

teddyrendahl:
1. This has been rebased on master to pass test with newest versions of `ophyd`
2. This how adds the plugins with `add_plugin`. This will be upset Travis until we receive a new PyDM tag

teddyrendahl:
Going to attack this from a different direction. Just going to start from a fresh branch.

PR #48, opened 2018-03-07
RunEngine Widget

### LINKED ISSUE #46: RunEngine Widget
## Expected Behavior

Basci RunEngine control. 
* Start / Stop / Pause
* Displays current state

teddyrendahl:
## Description

Basic Widget level control of the RunEngine. This includes:

* A colored label that indicates the current state of the Engine
* A dropdown QComboBox that presents the acceptable actions for the current engine state.
* All performed with signals and slots so other widgets can connect to RunEngine information without writing custom Python code.

### Remaining Questions
* Unfortunately if you expect plans to be run multiple times you need to provide a function that returns a plan not just a plan. This is because after the first run the generator will be exhausted. I couldn't quickly think of anyway around this but open to suggestions right now this looks likes:

```python
EngineWidget(plan=lambda: count(...))
```

## Motivation and Context

Closes #46 

## Screenshots
engine

### DIFF
--- a/typhon/engine.py
+++ b/typhon/engine.py
@@ -86,7 +86,7 @@ class QRunEngine(QObject, RunEngine):
         # Load the function from registry
         try:
             func = self.command_registry[command]
-        # Catch commands that we have no idea how to obey 
+        # Catch commands that we have no idea how to obey
         except KeyError as exc:
             logger.exception('Unrecognized command for RunEngine -> %s',
                              exc)
@@ -120,11 +120,11 @@ class EngineLabel(QLabel):
         color = self.color_map[state]
         self.setStyleSheet('QLabel {background-color: %s}' % color)
 
-
     def connect(self, engine):
         """Connect an existing QRunEngine"""
         engine.state_changed.connect(self.on_state_change)
         self.on_state_change(engine.state, None)
+
 
 class EngineControl(QStackedWidget):
     """
--- a/typhon/engine.py
+++ b/typhon/engine.py
@@ -10,18 +10,92 @@ import logging
 from bluesky import RunEngine
 from bluesky.utils import RunEngineInterrupted, install_qt_kicker
 from pydm.PyQt.QtGui import QVBoxLayout, QLabel, QComboBox, QGroupBox
-from pydm.PyQt.QtCore import pyqtSignal, pyqtSlot
+from pydm.PyQt.QtGui import QStackedWidget, QPushButton
+from pydm.PyQt.QtCore import QObject, pyqtSignal, pyqtSlot
 
 
 logger = logging.getLogger(__name__)
 
 
-def no_plan_warning(*args, **kwargs):
-    """
-    Convienence function to raise a user warning
-    """
-    logger.critical("Attempting to use the RunEngine "
-                    "without configuring a plan")
+class QRunEngine(QObject, RunEngine):
+
+    state_changed = pyqtSignal('QString', 'QString')
+    update_rate = 0.02
+    command_registry = {'Halt': RunEngine.halt,
+                        'Abort': RunEngine.abort,
+                        'Resume': RunEngine.resume,
+                        'Pause': RunEngine.request_pause,
+                        'Stop': RunEngine.stop}
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Attach the state_hook to emit signals
+        self.state_hook = self.on_state_change
+        # Create a kicker, not worried about doing this multiple times as this
+        # is checked by `install_qt_kicker` itself
+        install_qt_kicker(update_rate=self.update_rate)
+        # Allow a plan to be stored on the RunEngine
+        self.plan_creator = None
+
+    def on_state_change(self, state, old_state):
+        """
+        Report a state change of the RunEngine
+
+        This is added directly to the `RunEngine.state_hook` and emits the
+        `engine_state_change` signal.
+
+        Parameters
+        ----------
+        state: str
+
+        old_state: str
+
+        """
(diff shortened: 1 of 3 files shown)


teddyrendahl:
This should be re-evaluated in light of recent discussion about `RunEngine` as a `QObject`

ZLLentz:
I'm 50/50 on how the widget is set up. I like that it takes two clicks to start the `RunEngine`, I dislike that it takes two clicks to stop it. I'm not sold on the combo box, even though it's future-proof and makes the display very general.

Aesthetically it reminds me a lot of the lcls1 gui.

ZLLentz:
I think the widget is fine as is for a first pass, but we'll want to think about how to optimize the layout. I'm not going to tell you how to do this because I don't have any real suggestions on how to do it better.

teddyrendahl:
No I need feedback! I'm by in no means in love with the combobox. I might wait to merge and have a RunEngine `QObject`

teddyrendahl:
Should it be a stop button or a pause?

ZLLentz:
I suppose it should be pause, and then you can evaluate whether to resume or abort with the combobox.

klauer:
Agreed

ZLLentz:
Also, you should be able to use `itertools.tee` to run plans a second time, so you don't have to pass the empty lambda into the constructor.

typhon/engine.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+class EngineWidget(QGroupBox):
+    """
+    RunEngine Control Widget
+
+    Parameters
+    ----------
+    engine : RunEngine, optional
+        The underlying RunEngine object. A basic version wil be instatiated if
```
klauer: instatiated = in-satiated + instantiated? ;)

typhon/engine.py
```diff
@@ -0,0 +1,226 @@
...
+        # Do not allow plans to be set while RunEngine is active
+        if self.engine and self.engine.state != 'idle':
+            logger.exception("Can not change the configured plan while the "
+                             "RunEngine is running!")
+            return
+        # Store our plan internally
+        self._plan = plan
+        # Register a new call command
+        self.command_registry['Start'] = (lambda x:
+                                          RunEngine.__call__(x, self.plan()))
```
klauer: If I understand this correctly, you have to touch the right properties in the right order (engine -> plan), which modifies the command_registry dict, making up a new function to call the plan when 'Start' is selected.

This seems a bit over-complicated...  How about instead registering 'Start' as an instance method on the widget? Failing that, special-casing in 'command' below?

teddyrendahl:
Why is it desirable to have start be two clicks?

What if the QCombobox switches to a pushbutton when there is only one option?

typhon/engine.py
teddyrendahl: Interesting. I'll think about this. I also just realized that this setup enforces an order of operations in setting the plan and the engine. This absolutely needs to be corrected.

ZLLentz:
I don't want to accidentally start a plan, which is much more likely with a single click. This is an issue with `skywalker` where there is a chance you'll run a complicated procedure that wasn't desired when you really just wanted to show off your cool screen.

teddyrendahl:
Still have some documentation to do but I reworked this into a better state.

### Major Changes
* The `QComboBox` turns into a `QPushButton` when idle and running. It doesn't really make sense to me to have a `QComboBox` when there is only one option. (I am ignoring that you can halt  the RunEngine, you have to pause -> halt). This ignores @ZLLentz double click request. I think he has a valid point but this requires some thought and probably some use before I can think of a solution.
* All the RunEngine signals and slots are included on a new subclassed `QRunEngine`. This was inspired by some conversations with @tacaswell while he was here. Maybe this is something that belongs in `bluesky`? 
* Fixed the order of operations issue that @klauer pointed out. The `plan_creator` function is attached directly to `QRunEngine` so there is no confusion there.  There is still some kind of strange behavior where you make the widget with an engine and a plan, then swap the RunEngine, you won't have the same plan attached. I thought about automatically attaching the stored plan to the new engine, I just figured that I would rather have people explicitly add plans again instead of surprising folks by running a plan they didn't expect

### Screenshot
engine_v2

typhon/engine.py
```diff
@@ -0,0 +1,227 @@
...
+        ----------
+        state: str
+
+        old_state: str
+
+        """
+        self.state_changed.emit(state, old_state)
+
+    @pyqtSlot()
+    def start(self):
```
tacaswell: maybe require that the start slot take in a plan? That moves plan selection logic out to the containing widget and lets you do things like a drop-down list, full UIs for plans or a text box to just type the plan in.

tacaswell:
and :+1: this should be in bluesky, I suggest a new top-level module (which might eventually turn into a sub-package) `qt.py`.

typhon/engine.py
teddyrendahl: 👍 to this idea. This simplifies the concerns about attaching plan. It will require a little more work for users though. It probably means most applications will have to have do something like:

```python
but  = QPushButton('My Plan')
but.clicked.connect(functools.partial(QRE.start, my_plan()))
```

or in your example look up the plan from a dictionary based on a string. This is maybe slightly awkward but a lot more flexible. I can always wrap this up in a "PlanButton" or something. Also is there a point to having this by a `pyqtSlot`, can't imagine a lot of buttons will be emitting plans. We probably want to enforce this is a one to one relationship.

tacaswell: plans are all standard python objects, I don't see too much trouble passing live generators across a slot (the c++ layer should just pass it along as an opaque object).

tacaswell:
See https://github.com/nsls-ii/bluesky-ui for where related effort should land.

PR #50, opened 2018-04-06
API: Remove base Panel class

teddyrendahl:
## Description

In the first iteration of the UI all the panels had a `toggle` button that allowed things to be hidden. This worked alright but ended up being a large space waste. For the sake of cleanliness many of these should be combined into tabs and menus. 

This PR:

* Renames `Panel` to `TogglePanel` and moves it to the base `widget` file
* `SignalPanel` does not inherit from `TogglePanel`
* The `Configuration` and `Miscellaneous` signals are in a tab widget.
* Cleans component attribute names of their parent so -> xyzstage_x_stage -> x_stage 

## Motivation and Context

Part of a larger re-imagination of the user interface.

## Documentation
Updated Sphinx docs

## Screenshots
screen shot 2018-04-05 at 6 22 35 pm


klauer:
Neat! 👍

PR #52, opened 2018-05-14
UI: Clean grid layout

teddyrendahl:
## Description

Two minor changes to the grid setup.

1) Subclass `PyDMLineEdit` to set the `sizeHint`. Unfortunately this must be done on the class. 
2) Have the layout share the space between the `PyDMLineEdit` and `PyDMLabel` nicely.

## Screenshots
screen shot 2018-05-14 at 4 02 55 pm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,8 @@ before_install:
 
 install:
   - sudo apt-get update
+  # Install windows manager
+  - sudo apt-get install -y xvfb herbstluftwm
   # We do this conditionally because it saves us some downloading if the
   # version is the same.
   - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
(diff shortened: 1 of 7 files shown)


teddyrendahl:
Do not merge yet: PyDMComboBox still does not play nice.
screen shot 2018-05-15 at 4 57 15 pm

teddyrendahl:
I fell down a very deep rabbit hole but I can now see the light again. This is ready for review. There were two issues:

1) The TabWidget wanted all the tabs to have the same size widget. There is now an `add_tab` function that adds a given widget to the tab and makes sure that there is an expanding spacer so that the provided widget maintains its size.
2) `PyDMEnumComboBox` is contained within a layout which has some margins around it. This made those rows in the grid wider than necessary. This is now patched with `setContentsMargins(0, 0, 0, 0)` but that line can be deleted once https://github.com/slaclab/pydm/issues/315 is resolved.

Here is a screen shot of what the grid looks like now as well as a little sneak peek of what the stylesheet I am working on looks like:
grid_style

PR #61, opened 2018-05-31
ENH: Typhon Stylesheet

### LINKED ISSUE #57: Create QSS file
## Expected Behavior

Typhon will have a custom stylesheet. This should be specfied in `qss` format and will live in the `ui` directory.

teddyrendahl:
## Description

Formalized loading of the custom Typhon stylesheet I've been assembling over the last few months. This by no means the last iteration, but I wanted to make sure we are placing this in a consistent place within the library and distributing it regardless of packaging choice.

* Use `conda-forge` version of `ophyd`
* Added a `style.qss` file in `typhon/ui`
* Added a helper function to load the stylesheet
* Update `MANFIEST.in` to store `ui` and `qss` files
* Test that we have all functionality when package is not installed from source. Tests are now done twice; once from source, once from install to check packaging instructions.

## Motivation and Context

Closes #57 

## How Has This Been Tested?

We now run tests with the stylesheet by default. Also explicitly check the `load_stylesheet` function.

Added stylesheet instructions in Basic Usage Section

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,6 +76,12 @@ script:
   - coverage run run_tests.py --timeout 15 
   - coverage report -m
   - flake8 typhon
+  # Test again but with installed version
+  - mkdir notest
+  - mv typhon/*.* notest
+  - python run_tests.py
+  - mv notest/* typhon
+  - rmdir notest
   # Build docs
   - set -e
   - |
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,7 +48,6 @@ install:
   - conda install conda-build anaconda-client
   - conda update -q conda conda-build
   - conda config --append channels $PYDM_CHANNEL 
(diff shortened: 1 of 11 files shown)


ZLLentz:
Seems reasonable to me (not much to add), but I have to ask why you picked `application.setStyleSheet(typhon.load_stylesheet())` as your user-facing API for setting this instead of doing `app = QCoreApplication.instance()` in some other function and doing the rest for the user e.g.:

```
import typhon
from my_module import setup_my_app()

setup_my_app()
typhon.install_stylesheet()
```

 I'm also not sure what the preference is for "opt-in" vs "opt-out" in setting up the stylesheet.

The packaging all seems fine.

teddyrendahl:
That is a good point! I like that.

PR #62, opened 2018-05-31
ENH: SignalPlugin

teddyrendahl:
## Description

This should look very familiar. The last attempt did not take into account that we never want to recreate `ophyd` signals in the new `PyDMConnection` object. This new approach abandons the idea that we would every want to support the `SignalPlugin` from the `QtDesigner`, instead you register all the signals that you will need to create links for; then create the widgets that reference those signals.

```python
syn = ophyd.SynSignal(..., name='test_signal')
register_signal(syn)
widget  = PyDMWidget('sig://test_signal')
```

This is a little cumbersome but in reality most users will not be calling these by themselves. When they create a `DeviceDisplay` we can pass through their device and register signals for them.  

**Note** This won't pass all the tests until a new tag of PyDM is created

## Motivation and Context

Closes #47 

## How Has This Been Tested?

Tested with a `PyDMWritableWidget`

## Documentation
Still lacking. Will fill this in if the approach is considered valid.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,6 +76,12 @@ script:
   - coverage run run_tests.py --timeout 15 
   - coverage report -m
   - flake8 typhon
+  # Test again but with installed version
+  - mkdir notest
+  - mv typhon/*.* notest
+  - python run_tests.py
+  - mv notest/* typhon
+  - rmdir notest
   # Build docs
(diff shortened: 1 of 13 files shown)


teddyrendahl:
Wrong branch.

Issue #17, opened 2017-11-08
Save generated user interfaces


teddyrendahl:
## Expected Behavior

Save the automatically generated user interfaces to disk. This allows for later access without having to instantiate an `ophyd` devices.
## Possible Solution

Since the `Panel` class rips out the `PV` names and uses them directly, these can be saved as is. However, all the function references will need to be wiped clean. Later, we could imagine a way that an `ophyd` device can be reinstantiated (maybe through a `subclass` of TyphonApplication`. However, this not necessary for the first prototype

teddyrendahl:
De-scoping from `0.1.0` this should be done with the built-in `PyDMDisplay`. I'm not sure it is feasible to save `.ui` file and have a user by hand.

teddyrendahl:
Maybe we save a `JSON` representation not a `.ui` file and `Typhon` learns to recreate itself...

PR #63, opened 2018-05-31
ENH: SignalPlugin

### LINKED ISSUE #28: Allow Python properties to be be shown in the UI
## Expected Behavior

A widget that was capable of polling a Python variable and broadcasting changes to the screen.  Just like we currently support `add_pv` and `add_signal`, we could support `add_property`. This would not be done automatically, but would be specified by the user similar to methods.

## Current Behavior

Not supported

## Possible Solution

The simplest solution seems to make a PyDM py:// plugin if it does not exist.

### LINKED ISSUE #44: SignalPlugin

teddyrendahl:
## Description

This should look very familiar. The last attempt did not take into account that we never want to recreate `ophyd` signals in the new `PyDMConnection` object. This new approach abandons the idea that we would every want to support the `SignalPlugin` from the `QtDesigner`, instead you register all the signals that you will need to create links for; then create the widgets that reference those signals.

```python
syn = ophyd.SynSignal(..., name='test_signal')
register_signal(syn)
widget  = PyDMWidget('sig://test_signal')
```

This is a little cumbersome but in reality most users will not be calling these by themselves. When they create a `DeviceDisplay` we can pass through their device and register signals for them.  

**Note** This won't pass all the tests until a new tag of PyDM is created

## Motivation and Context

Closes #28 
Closes #44 

## How Has This Been Tested?

Tested with a `PyDMWritableWidget`

## Documentation
Still lacking. Will fill this in if the approach is considered valid.

### DIFF
--- a/docs/source/connections.rst
+++ b/docs/source/connections.rst
@@ -1,0 +1,33 @@
+=======================
+Application Connections
+=======================
+
+Ophyd Signals
+=============
+Typhon takes advantage of the flexible data plugin system contained within
+``PyDM`` and the abstraction of the "control layer" within ``Ophyd``. In the
+:class:`.SignalPanel`, objects signals are queried for their type. If these are
+determined to be coming from ``EPICS`` the data plugin configured within
+``PyDM`` is used directly, any other kind of signal goes through the generic
+:class:`.SignalPlugin`. This uses the subscription system contained within
+``Ophyd`` to keep widgets values updated. One caveat is that ``PyDM`` requires
+that channels are specified by a string identifier. In the case of
+``ophyd.Signal`` objects we want to ensure that these are passed by reference
+to avoid duplicating objects. This means the workflow for adding these has one
+more additonal step where the ``Signal`` is registered with the ``PyDM``
+plugin.
+
+.. code:: python
+
+   from typhon.plugins import register_signal
+
+   # Create an Ophyd Signal
+   my_signal = ophyd.Signal(name='this_signal')
+   # Register this with the Plugin
+   register_signal(my_signal)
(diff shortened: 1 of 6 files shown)


ZLLentz:
I haven't looked line-by-line yet, but I like the approach. One concern: suppose I want to take a `typhon` generated screen and modify it, but keep it valid for all objects of the same class. Would I have to go through all of the generated `sig://` channels and figure out where to cut off the name prefixes? Though even then, I'm not up to date on the status of templating in `pydm`.

teddyrendahl:
@ZLLentz very valid point. I think there is way I could sub in the device name as a macro so it could be switched out slightly earlier, however right now `Typhon` will only create displays for objects not classes. There are a few issues that revolve around this largely #9 and #17. I think I might punt on this and come back through when I address those.

teddyrendahl:
Various issues regarding typing. I decided the implementation most faithful to `ophyd` is to trust the `dtype` value given in `describe` then enforce strictly that our values match that while avoiding raising exceptions.

teddyrendahl:
Merging this. Too many things waiting for this downstream to wait for PyDM tag.

PR #91, opened 2018-07-12
MAINT: Catch exceptions created on Subscription

teddyrendahl:
## Description

Ran into an `ophyd.Signal` that raised an exception upon subscription.  The `SignalPlugin` did not wrap `subscribe` up in a `try / except` block which brought everything to a screeching halt.

Now, the solution is not quite so simple as just adding the `try` block in `SignalConnection`. If we raise an error on subscription we **want** that `SignalConnection` to explode. Otherwise we run the risk of subscribing a bunch of channels to a connection that will never update! So instead my solution was to add this to the `add_connection` block. If we fail to subscribe we make sure not to save the `SignalConnection` and the next channel that needs the `signal` will try again.

@hhslepicka Does this belong in PyDM? Or are `PyDMPlugin` objects responsible for catching their own kind of exceptions?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-sudo: false
+sudo: required
 
 env:
   global:
--- a/typhon/plugins/core.py
+++ b/typhon/plugins/core.py
@@ -175,3 +175,15 @@ class SignalPlugin(PyDMPlugin):
     """Plugin registered with PyDM to handle SignalConnection"""
     protocol = 'sig'
     connection_class = SignalConnection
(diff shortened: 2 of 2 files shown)


ZLLentz:
I'm not sure I fully understand the nuances of the problem, but I'm in favor of not crashing the GUI on subscription exceptions.

teddyrendahl:
@hhslepicka My worry is that different plugins will want different rules i.e keep my connection object and just try adding another listener or create a whole new connection e.t.c

PR #92, opened 2018-07-13
ENH: Signal Metadata

### LINKED ISSUE #85: SignalPlugin should handle enum_strs
## Expected Behavior

Generic Ophyd signals can have `enum_strs` as well. We should make sure we send these to PyDM to make proper use of `QComboBox` e.t.c

teddyrendahl:
## Description

Some general improvements to the `SignalPlugin`

* We now inspect the `describe` dictionary of the `Signal` for `"enum_strs"`, `"units"` and `"precision"`. I also left this pretty extensible so that in the future as long as we have a describe key that maps to a signal it is just a matter of adding a tuple. 
* Very much the successor to #91, there was the potential for exceptions on `get` when setting up our signal that would not have been caught. We now wrap these up nicely. Also now that `SignalPlugin` handles some of the error handling I moved more things there. It makes more sense to not create a `SignalConnection` at all if you give me the name of a non-existant `Signal`.

### Concerns
There is still the potential for a few strange edge cases to arise if our network is unstable. For instance if we successfully subscribe to a signal, but then fail to get it e.t.c. I can try and armor this Plugin up against all these situations but the real root of the issue is that the base `Signal` has not callback for connection. I'm kind of left guessing if my last `get` failed because of a hiccup or because my signal is really gone.

Also @hhslepicka I am assuming I don't have to be diligent about sending the connection state to `False`. Is this true? Widgets will default to their disconnected state? I ask because I don't want one bad `get` to mark every listener as disconnected....

## Motivation and Context

Closes #85 

## How Has This Been Tested?

* Added a test to send `metadata` through describe
* Add a test with a signal that raises `TimeoutError`

- [x] Add notes on how signal picks up metadata

### DIFF
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -20,6 +20,13 @@ class RichSignal(Signal):
                              'shape': []}}
 
 class DeadSignal(Signal):
+    subscribable = False
+
+    def subscribe(self, *args, **kwargs):
+        if self.subscribable:
+            pass
+        else:
+            raise TimeoutError("Timeout on subscribe")
 
     def get(self, *args, **kwargs):
         raise TimeoutError("Timeout on get")
@@ -61,14 +68,6 @@ def test_signal_connection(qapp):
     qapp.processEvents()
     assert sig.get() == 3
 
-def test_invalid_signal():
-    widget = WritableWidget()
-    listener = widget.channels()[0]
-    # Invalid Signal
-    sig_conn = SignalConnection(listener, 'my_signal')
-    assert not widget._connected
-    assert not widget._write_access
-
 
 def test_metadata(qapp):
     widget = WritableWidget()
@@ -87,10 +86,16 @@ def test_metadata(qapp):
 def test_disconnection(qapp):
     widget = WritableWidget()
     listener = widget.channels()[0]
-    # Create a signal and attach our listener
+    listener.address = 'sig://invalid'
+    plugin = SignalPlugin()
+    # Non-existant signal doesn't raise an error
+    plugin.add_connection(listener)
+    # Create a signal that will raise a TimeoutError
     sig = DeadSignal(name='broken_signal', value=1)
     register_signal(sig)
+    listener.address = 'sig://broken_signal'
+    # This should fail on the subscribe
+    plugin.add_connection(listener)
+    # This should fail on the get
+    sig.subscribable = True
     sig_conn = SignalConnection(listener, 'broken_signal')
-    sig_conn.add_listener(sig_b)
-    assert not widget._connected
-    assert not widget._write_access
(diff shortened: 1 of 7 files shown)


hhslepicka:
@teddyrendahl... the widget will be by default disconnected until you tell it the opposite.
Once it is `connected` it will remain like that until a signal is emitted saying that the channel is no longer connected. Does that answer the question?

tests/test_plugins.py
```diff
@@ -10,6 +10,31 @@ class WritableWidget(QWidget, PyDMWritableWidget):
     pass
 
 
+class RichSignal(Signal):
```
hhslepicka: 💰

tests/test_plugins.py
```diff
@@ -10,6 +10,31 @@ class WritableWidget(QWidget, PyDMWritableWidget):
...
+
+    def describe(self):
+        return {self.name : {'enum_strs': ('a', 'b', 'c'),
+                             'precision': 2,
+                             'units': 'urad',
+                             'dtype': 'number',
+                             'shape': []}}
+
+class DeadSignal(Signal):
```
hhslepicka: 💀

typhon/plugins/core.py
```diff
@@ -181,9 +184,13 @@ def add_connection(self, channel):
         try:
             # Add a PyDMConnection for the channel
             super().add_connection(channel)
-        # There is a chance that we raise an Exception on subscription. If so,
-        # don't add this to our list of good to go exceptions so the next
+        # There is a chance that we raise an Exception on creation. If so,
+        # don't add this to our list of good to go connections. The next
         # attempt we try again.
```
hhslepicka: What do you mean by: "The next attempt we try again?
If you have a widget asking to connect to a certain plugin and it fails I don't think that it will request the connection again... maybe we should add something at PyDM to retry? I'm not sure if I ever saw this case.
For Epics we add the listener anyway since we get a notification when the PV is connected... In this case, you are not adding the connection to the channels when it fails to subscribe...

teddyrendahl:
@hhslepicka That does answer the question. Just wanted to make sure I wasn't making a wrong assumption. I won't worry about sending `connection_state.emit(False)` too much then.

typhon/plugins/core.py
teddyrendahl: I mean, if you have two listeners request the same channel, the first one fails, we still will try again on the second i.e

```python
PyDMWidget(init_channel='sig://my_signal')  # Fails from network hiccup
PyDMWIdget(init_channel='sig://my_signal')  # Try again to create a channel
```

teddyrendahl:
Realized that I needed to teach `Typhon` to draw `QCombobox` for the plain `Signal` connections as well. Found that I wasn't wrapping `enum_strs` in `try/except` there either so reworked that.

Issue #107, opened 2018-10-06
Repetitive Tool Loading


teddyrendahl:
The current incarnation creates an entire new `DeviceDisplay` for each component device then hides the toolbar. The issue is that tools are loaded when the device screen is created, meaning we take the brunt of loading a ton of screens we will never use.  The tools need to be adapted to be one tool for `TyphonDisplay` not one per device

PR #109, opened 2018-10-06
MAINT: Tool Restructure

### LINKED ISSUE #106: Refactor Devices with only `__init__` to have classmethods
Many of the tools follow the format "Typhon_} which starts empty, then "Device_" which autopopulates the widget with device information.

I'm proposing the general interface.

```python
class TyphonTool:
   
     def add_device(self, device):

     @classmethod
      def from_device(cls, device):
            foo = cls()
            foo.add_device(device)
            return cls
```

Deprecation warnings will be issued for old API

teddyrendahl:
## Description

Created a base API for all `typhon` tools to follow. This makes official the new structure in which every tool can be instantiated by itself, but if you want to automatically make it useful with an object you can use a `from_device` class method instead.

Also made a new widget `TyphonLogDisplay` to contain the `PyDMLogDisplay` widget. This was necessary to support multiple devices logging to the same widget. It also leaves room to expand the feature set of the widget to save the log, post to elog e.t.c
 
## Motivation and Context

Closes #106 
Part way to addressing #107 

## How Has This Been Tested?

Added test for `TyphonLogDisplay`

Documentation has been ignored for too long. Addressing this in the following PR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ logs/*
 __pycache__/
 *.py[cod]
 *$py.class
+*.pytest_cache
 
 # Ignore generated MacOS files
 .DS_Store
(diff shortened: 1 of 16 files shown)


teddyrendahl:
Got inspired. Documentation rectified!

PR #116, opened 2018-10-11
TST: Qtbot

teddyrendahl:
## Description

Add `qtbot` to all tests where widgets are created.

## Motivation and Context

I hate segfaults, you hate segfaults, we all hate segfaults

### DIFF
--- a/tests/test_log.py
+++ b/tests/test_log.py
@@ -5,9 +5,10 @@ from ophyd import Device
 from typhon.tools import TyphonLogDisplay
 
 
-def test_log_display(qapp):
+def test_log_display(qapp, qtbot):
     dev = Device(name='test')
     log_tool = TyphonLogDisplay.from_device(dev)
+    qtbot.addWidget(log_tool)
     dev.log.error(dev.name)
     assert dev.name in log_tool.logdisplay.text.toPlainText()
     dev2 = Device(name='blah')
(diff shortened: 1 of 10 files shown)


tests/test_log.py
```diff
@@ -5,9 +5,10 @@
 from typhon.tools import TyphonLogDisplay
 
 
-def test_log_display(qapp):
+def test_log_display(qapp, qtbot):
```
hhslepicka: `qapp` is not needed in here...

tests/test_plugins.py
```diff
@@ -73,8 +75,9 @@ def test_metadata(qapp):
     assert widget._prec == 2
 
 
-def test_disconnection(qapp):
+def test_disconnection(qapp, qtbot):
```
hhslepicka: `qapp` is not used in here...

teddyrendahl: 👍

tests/test_log.py
teddyrendahl: 👍

teddyrendahl:
@hhslepicka I tried to remove them and the tests failed. I'm guessing there is a better way to do it with `qtbot.waitUntil` but I kind of just want to move on.

hhslepicka:
Agreed!

PR #118, opened 2018-10-23
Happi Plugin

### LINKED ISSUE #38: Ophyd Plugin
## Expected Behavior

Create a PyDM Ophyd Plugin

## Current Behavior

No ophyd plugin available

## Possible Solution

Specify an ophyd device class, args and kwargs in the channel anme then create the DeviceDisplay with the proper instantiated device behind it.

teddyrendahl:
## Description

A `PyDMPlugin` that can transmit objects and their metadata. This uses the proposed model that will be used in `PyDM` for structured data in the future. There is one signal that sends a dictionary of information to subscribers and it is up to them to parse it. In this case it includes the instantiated object (usually an `ophyd.Device`) and the `happi.Device` that it came from.

Thoughts
-------
This does the job of getting devices to widgets but is kind of awkward in some ways. For instance, if another widget requests the same device, we don't want to send out that information again (might be more expensive to redraw entire devices). In fact we really only want to send the device once, I hope I'm not slamming a round peg in a square hole here

## Motivation and Context

Closes #38 

## How Has This Been Tested?

Added tests for adding `HappiConnection`

- []

### DIFF
--- a/typhon/plugins/happi.py
+++ b/typhon/plugins/happi.py
@@ -1,6 +1,7 @@
 import logging
 
-from happi import Client, from_container
+from happi import Client
+from happi.loader import from_container
 from happi.errors import SearchError
 from pydm.data_plugins.plugin import PyDMPlugin, PyDMConnection
 from pydm.widgets.channel import PyDMChannel
--- a/typhon/plugins/happi.py
+++ b/typhon/plugins/happi.py
@@ -1,0 +1,88 @@
+import logging
+
+from happi import Client, from_container
+from happi.errors import SearchError
+from pydm.data_plugins.plugin import PyDMPlugin, PyDMConnection
+from pydm.widgets.channel import PyDMChannel
+from qtpy.QtCore import Signal, Slot
+
+_client = None
+logger = logging.getLogger(__name__)
+
+
+def register_client(client):
+    """Register a Happi Client to be used with the DataPlugin"""
+    global _client
+    _client = client
+
+
+class HappiChannel(PyDMChannel):
+    """
+    PyDMChannel to transport Device Information
+
+    Parameters
+    ----------
+    tx_slot: callable
+        Slot on widget to accept a dictionary of both the device and metadata
+        information
+    """
+    def __init__(self, *, tx_slot, **kwargs):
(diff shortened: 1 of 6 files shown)


teddyrendahl:
Someone mind taking a look at this?

typhon/plugins/happi.py
```diff
@@ -0,0 +1,90 @@
...
+        # If we haven't made a Client by the time we register the Plugin. Try
+        # and load one from configuration file
+        if not _client:
+            register_client(Client.from_config())
+
+    def add_connection(self, channel):
+        try:
+            super().add_connection(channel)
+        except SearchError:
+            logger.exception("Unable to find device in happi database.")
```
klauer: `SearchError` just means not found? I'd lean toward `error` instead of seeing a traceback for this.

```suggestion
            logger.error("Unable to find device in happi database: %s", ex)
```

teddyrendahl: `pcds-utils`!

ZLLentz: I'll have you know I got five minutes into starting a `pcds_utils` module last week before getting distracted

teddyrendahl:
Kicking to test again with https://github.com/slaclab/pydm/pull/355 merged...

typhon/plugins/happi.py
klauer: 😄 Yeah, that's a better pattern.

Though I think there should be in happi a clear distinction between failures: `OhCrapSearchWentHorriblyWrong` vs `NoMatchingDeviceFound`

For the former, give me the traceback. For the latter, stop wasting my screen real estate!

ZLLentz: I agree with Ken, `happi` should have higher error granularity

teddyrendahl: I think that is what `SearchError` does.

teddyrendahl:
Going to rebase after #120 changed the testing suite. Also hoping to do similar work to make sure testing suite passes on past and present PyDM versions

teddyrendahl:
Fixed logging, added a test for the disconnect which uncovered a bug. This should now be good.

PR #120, opened 2018-10-30
TST: SignalConnection 

teddyrendahl:
## Description

After https://github.com/slaclab/pydm/pull/355, `PyDMChannel` will connect itself.  This caused some issues not in the code itself, but the testing suite where we monkeyed around connecting signals to plugins itself. Also was adding some `print` statements to discombobulate what was going on and realized that some more logging messages would be helpful. 😄

### DIFF
--- a/tests/test_plugins.py
+++ b/tests/test_plugins.py
@@ -31,6 +31,7 @@ def test_signal_connection(qapp, qtbot):
     register_signal(sig)
     widget = WritableWidget()
     qtbot.addWidget(widget)
+    widget.channel = 'sig://my_signal'
     listener = widget.channels()[0]
     sig_conn = SignalConnection(listener, 'my_signal')
     sig_conn.add_listener(listener)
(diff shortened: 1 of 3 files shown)


tests/test_plugins.py
```diff
@@ -31,9 +31,13 @@ def test_signal_connection(qapp, qtbot):
...
     qtbot.addWidget(widget)
+    widget.channel = 'sig://my_signal'
     listener = widget.channels()[0]
-    sig_conn = SignalConnection(listener, 'my_signal')
-    sig_conn.add_listener(listener)
+    # If PyDMChannel can not connect, we need to connect it ourselves
+    # In PyDM > 1.5.0 this will not be neccesary as the widget will be
+    # connected after we set the channel name
+    if not hasattr(listener, 'connect'):
+        qapp.establish_widget_connections(widget)
```
hhslepicka: 👍

tests/test_plugins.py
```diff
@@ -49,7 +53,7 @@ def test_signal_connection(qapp, qtbot):
     # Try changing types
     qapp.processEvents()
     qapp.processEvents()  # Must be called twice. Multiple rounds of signals
-    sig_conn.remove_listener(listener)
+    qapp.close_widget_connections(widget)
```
hhslepicka: Maybe it would be wise to also wrap this with the check for the `hasattr`?
We will deprecate the `close_widget_connections` at the qapp in a couple of releases.
For now it will work but it can start failing later on.

teddyrendahl: 👍

PR #122, opened 2018-11-08
TyphonPanel

### LINKED ISSUE #39: Typhon Designer Plugin
## Expected Behavior

Create a Typhon Designer Plugin so that the DeviceDisplay can be used in a drag and drop editor.

## Current Behavior

Widgets are not available in Designer

## Possible Solution

Typhon Designer Plugin that takes an Ophyd plugin specification for each widget

teddyrendahl:
## Description

A generalized widget to show a group of signals on a `Device`.  Specify a channel with a `happi` address and select which `Kind` of signals you want! 

I still need to setup the machinery to add the designer plugins on your path, but I want to think a little bit about the cleanest way to do this.

## Motivation and Context

Closes #39 

## How Has This Been Tested?

Added a test for panel creation and `Kind` selection

designer


typhon/signal.py
```diff
@@ -178,3 +184,98 @@ def _add_row(self, read, name, write=None):
...
+        # We can't set a layout if we don't have any devices
+        if not self.devices:
+            return
+        # Clear our layout
+        self.layout().clear()
+        shown_kind = [kind for kind in Kind if self._kinds[kind.name]]
+        # Iterate through kinds
+        for kind in (Kind.hinted, Kind.normal, Kind.config, Kind.omitted):
+            if kind in shown_kind:
+                for (attr, signal) in grab_kind(self.devices[0], kind.name):
```
klauer: Do we want signals grouped by kind or by attribute here?

teddyrendahl: 🤔  My thought was you always want the higher priority signals at the top

klauer: Sort order option? ;)

teddyrendahl:
* Some various `flake8` cleanup because apparently there were more checks added with regards to unused variables
* Made back-compatible with older versions of PyDM where `PyDMChannel` can't connect itself.

typhon/signal.py
teddyrendahl: I think this is correct. No right or wrong answer here and not hard to support both

teddyrendahl:
Added sorting and blocked some base `pyqtProperty` stuff from `PyDM`.

PR #128, opened 2018-11-26
ENH: TyphonDisplay in Designer

### LINKED ISSUE #101: Embedded Screen
## Expected Behavior

Auto-generate an embeddable display. This should be just a condensed version of the `TyphonDisplay`

### LINKED ISSUE #115: Formalize Template Declaration
## Expected Behavior

Typhon should be as clear as possible about separating code from templates. Further yet, we should have clear documentation about what every class expects to have in inside its template. This would mean that each object should have a template that is then automatically filled by specific devices. Different devices could have different templates that specify more information (PVs) but are autofilled via the Typhon.

teddyrendahl:
## Description

Configure `TyphonDisplay` to:
* Be able to be used in designer with `HappiPlugin`
* Be able to load arbitrary templates

There are three kind of templates: `embedded`, `detailed` and `engineering`. You can control which one you want by using the `template_type` property. In this mode of operation, a device is added via the `channel` attribute or direct call of `add_device` in a Python session and the device metadata is inspected to see if it specifies a screen for this type. If none are found we use the default `Typhon` template. You can ignore any screen given in the metadata by checking the `use_default_templates` boolean. You can also ignore all of this machinery and just put an absolute path in `use_template`. This will be used regardless of the other settings.

## Motivation and Context

Closes #101 
Closes #115

## How Has This Been Tested?

Added tests for new templated behavior.

At this point, docs need an entire overhaul...

display


teddyrendahl:
Anyone want to review? Realize it is large but looking for some assurance on approach.

typhon/display.py
```diff
@@ -42,100 +58,141 @@ class TyphonDisplay(TyphonBase):
 
     parent: QWidget, optional
     """
-    def __init__(self, name=None, image=None, parent=None):
+    # Template types and defaults
+    Q_ENUMS(TemplateTypes)
```
ZLLentz: Is this why you need that confusing `TemplateTypes` class? Qt being weird

typhon/display.py
```diff
@@ -1,19 +1,35 @@
...
+    def to_enum(cls):
+        # First let's remove the internals
+        entries = [(k, v) for k, v in cls.__dict__.items()
+                   if not k.startswith("__")
+                   and not callable(v)
+                   and not isinstance(v, staticmethod)]
+        return Enum('TemplateEnum', entries)
+
+
+class TyphonDisplay(TyphonBase, TyphonDesignerMixin, TemplateTypes):
```
ZLLentz: I don't think this needs to subclass `TemplateTypes`

typhon/display.py
```diff
@@ -42,100 +58,141 @@ class TyphonDisplay(TyphonBase):
...
+        if self._use_template:
+            return self._use_template
+        # Search in the last macros, maybe our device told us what to do
+        template_key = self.TemplateEnum(self._template_type).name
+        if not self._use_default and self._last_macros.get(template_key, None):
+            return self._last_macros[template_key]
+        # Otherwise just use the default
+        return self.default_templates[template_key]
+
+    @Property(TemplateTypes)
```
ZLLentz: What does capital P property do? Is this another weird Qt thing? What an awful API

typhon/display.py
teddyrendahl: For stupid Qt it does. When this class gets saved in a `.ui` file it stores the parameters as `TyphonDisplay.Engineering` e.t.c

typhon/display.py
teddyrendahl: Yes ... this is terrible, but neccesary

typhon/display.py
teddyrendahl: This is what the `qtpy` renames `pyqtproperty` as

typhon/display.py
ZLLentz: That's awful

typhon/display.py
ZLLentz: That's awful

typhon/display.py
ZLLentz: That's awful

klauer:
Outside of the `qtpy.Property` thing, I'd be interested to investigate alternatives at some point, if there are any. No reason to hold up progress, though. 👍

hhslepicka:
It is really awful, nasty and a pain but unfortunately this is how folks get around with the properties to show up properly at Designer...
@klauer if you find an alternative please let us know!

PR #130, opened 2018-12-06
Designer Environment

### LINKED ISSUE #129: Standardize PYDMDESIGNER environment
- [ ] Create a shell script that setups the PYQTDESIGNERPATH and PYDMDESIGNERONLINE
- [ ] Source this when we activate the environment

teddyrendahl:
## Description

If installed via CONDA, `typhon` widgets will now automatically be available in Designer. For development, the `typhon_env.sh` script can be used to set everything up as well. The PYDM_DESIGNER_ONLINE variable is set up as well, otherwise the Typhon widgets will not update as you change parameters.

## Motivation and Context

Closes #129 

## How Has This Been Tested?

Test will be run if we believe the package to be installed via CONDA

### DIFF
--- a/bin/typhon_env.sh
+++ b/bin/typhon_env.sh
@@ -1,0 +1,5 @@
+#!/usr//bin/env bash
+TYPHON_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+export PYQTDESIGNERPATH="$TYPHON_DIR":$PYQTDESIGNERPATH
+export PYDM_DESIGNER_ONLINE=True
+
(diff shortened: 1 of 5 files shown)


bin/typhon_env.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr//bin/env bash
```
ZLLentz: Extra slash in the middle here

bin/typhon_env.sh
```diff
@@ -0,0 +1,5 @@
...
+TYPHON_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
```
ZLLentz: Is this the normal way to do this? Seems weird to cd and pwd to get the full path

teddyrendahl: I think we usually do `readlink`, but I wanted this to work on MacOS as well (which doesn't ship with `readlink`)

ZLLentz: is `dirname` not sufficient?

teddyrendahl: I like this... my only concern is that the result of the script might change based on directory. Let's say you have `typhon` installed in your environment, but you are testing a modified local checkout. Maybe I'm overthinking this.

teddyrendahl: I like yours better, slightly more flexible. Unfortunately the people who really care about the designer are often the least knowledgable... trying to make it as simple as possible.

PR #136, opened 2018-12-07
TyphonDisplay API

teddyrendahl:
Played around with the API of `TyphonDisplay` today, trying to put it in a `TyphonSuite`, and it really was awkward. I think this solution is a lot more intuitive. What you really want to do is not just specify a static template but specify a template for each kind of display (if you aren't already using `happi`). That way you still get the full feature set of swapping around the `TemplateType` but you can modify the behavior at whichever stage you deem necessary. For instance, the situation where you wanted a custom embedded template but wanted to use the other defaults was hard to specify in the old API.

## Changes
`use_template`is gone, instead there are just properties for each type `embedded_template`, `display_template` e.t.c. The template priority that is used is as follows:

1. If `use_default_templates` is forced we use the default template shipped with `Typhon`
2. If a template is provided in the corresponding property we use it
3. If a template is provided in the `happi` metdata, we use it.
4.  Otherwise, use the default.

Also, `image` and `methods` were removed in the PR but there were still explicit keywords for these in `TyphonSuite`. These have been removed.

### DIFF
--- a/tests/test_display.py
+++ b/tests/test_display.py
@@ -43,15 +43,17 @@ def test_device_display_templates(motor, qtbot):
     assert panel.current_template == panel.default_templates['embedded_screen']
     # Force a specific template
     eng_ui = panel.default_templates['engineering_screen']
-    panel.use_template = eng_ui
-    assert panel.use_template == eng_ui
+    panel.embedded_template = eng_ui
+    assert panel.embedded_template == eng_ui
     assert panel.current_template == eng_ui
     # Check that if we pass in a template as macros we use the forced template
     panel.load_template(macros={'embedded_screen': 'tst.ui'})
     assert panel.current_template == eng_ui
+    # But if we don't specify any template use macros
+    panel.embedded_template = ''
+    assert panel.current_template == 'tst.ui'
     panel.use_default_templates = True
     assert panel.use_default_templates
-    panel.use_template = ''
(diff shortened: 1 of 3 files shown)


typhon/display.py
```diff
@@ -59,19 +59,18 @@ class TyphonDisplay(TyphonBase, TyphonDesignerMixin, TemplateTypes):
...
     def __init__(self,  parent=None, **kwargs):
         # Intialize background variable
-        self._use_template = ''
-        self._use_default = False
+        self._forced_template = ''
         self._last_macros = dict()
         self._main_widget = None
-        self._template_type = TemplateTypes.detailed_screen
+        self._display_type = DisplayTypes.detailed_screen
+        self.templates = dict((_typ.name, os.path.join(ui_dir,
```
klauer: `ui_dir` -> `typhon_package_ui_path` or something?

tests/test_display.py
```diff
@@ -1,9 +1,20 @@
...
 from typhon.utils import clean_attr
+
 from .conftest import show_widget
 
 
+@pytest.fixture(scope='function')
+def display(qtbot):
+    display = TyphonDisplay()
+    qtbot.addWidget(display)
+    return display
```
klauer: 👍

typhon/display.py
teddyrendahl: hmmmmm, that is an awfully vague name

PR #137, opened 2018-12-07
Python 3.7

teddyrendahl:
Be ready!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ matrix:
   include:
      - python: 3.6
        env:
+          - BUILD_DOCS=1
           - BUILD=1
           - PYDM_CHANNEL=pcds-tag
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -10,10 +10,25 @@ matrix:
...
      - python: 3.6
        env:
+          - BUILD_DOCS=1
           - BUILD=1
           - PYDM_CHANNEL=pcds-tag
      - python: 3.6
        env: PYDM_CHANNEL=pydm-dev
+     - python: 3.7
+       dist: xenial
+       sudo: true
```
klauer: I think `sudo` should be required on all, given the `apt-get` calls below.

teddyrendahl: I have `sudo: true` at the top. Maybe this is just redundant

PR #138, opened 2018-12-07
FIX: Point to bin folder not typhon source

teddyrendahl:
Slight error in `typhon_env.sh` script from #130

### DIFF
--- a/bin/typhon_env.sh
+++ b/bin/typhon_env.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-TYPHON_DIR=$(python -c 'import typhon; import pathlib; print(pathlib.Path(typhon.__file__).parent)')
-export PYQTDESIGNERPATH="$TYPHON_DIR":$PYQTDESIGNERPATH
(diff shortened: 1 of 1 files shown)


ZLLentz:
Does this file belong in bin?

teddyrendahl:
Where would you place it? I guess you don't want to execute it you want to source it. Just put in a `scripts` directory?

klauer:
`etc`, `scripts`...?

ZLLentz:
Sorry! I should have been more specific. I meant `typhon_designer_plugin.py`. It's not a script you run or source in any sense.

ZLLentz:
e.g. maybe you need to move the python file, not change the directory you set in the environment variable

teddyrendahl:
So I can't point to the source because whatever disgusting thing that `QtDesigner` does to run that file craps out with relative imports. I guess that doesn't necessarily require that file living outside that directory though...

ZLLentz:
I still think it's structurally inconsistent to have non-executable python scripts in bin. It will be confusing when I look in bin to find any typhon scripts, and it will be confusing when I go looking for the qt designer plugins and find them in the executables directory. But I won't harp on this any more.

PR #140, opened 2018-12-10
Device Categories

### LINKED ISSUE #133: Support arbitrary device categories
The sidebar should support arbitrary device categories, not just put everything under "Devices"

teddyrendahl:
## Description

The goal is to be more flexible with how we add widgets to the sidebar. Before we had one category for "Devices" one for "Tools" and and that was it. In reality, if you get more than a few devices in the suite you may want to start breaking this up into more fine grain categories (below motion and diagnostics are separated). 

In order to accommodate this I had to change how we searched through the `ParameterTree` to find relevant object and keep the docking status inline with the tree button.

I also had deprecated the `add_subdevice` method. But in line with other PRs I don't want to keep that kind of flotsam around this early in the project so it is officially gone.

## Motivation and Context

Closes #133 

## How Has This Been Tested?

Revamped test suite for TyphonSuite 😄  Unit tests are more unit-like

Updated docstrings as well

categories

### DIFF
--- a/tests/test_suite.py
+++ b/tests/test_suite.py
@@ -46,25 +46,60 @@ def test_suite_tools(suite):
     assert len(suite.tools[0].devices) == 1
 
 
-def test_suite_subdisplay(qtbot, suite):
-    device = suite.devices[0]
-    x_display = suite.get_subdisplay(device.x)
-    assert device.x in x_display.devices
+def test_suite_get_subdisplay_by_device(suite, device):
+    display = suite.get_subdisplay(device)
+    assert device in display.devices
+
+
+def test_suite_get_subdisplay_by_name(suite, device):
+    display = suite.get_subdisplay(device.name)
+    assert device in display.devices
+
+
+def test_suite_show_display_by_device(suite, device):
     suite.show_subdisplay(device.x)
-    assert isinstance(x_display.parent(), QDockWidget)
-    # Set display by name
-    y_display = suite.get_subdisplay(device.y)
-    qtbot.addWidget(y_display)
-    assert device.y in y_display.devices
-    suite.show_subdisplay(clean_name(device.y))
-    assert isinstance(y_display.parent(), QDockWidget)
-    with qtbot.waitSignal(y_display.parent().closing):
-        # Hide all our subdisplays
-        suite.hide_subdisplays()
-    return suite
+    dock = suite.layout().itemAt(suite.layout().count() - 1).widget()
+    assert isinstance(dock, QDockWidget)
+    assert device.x in dock.widget().devices
 
 
-@show_widget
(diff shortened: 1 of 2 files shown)


tests/test_suite.py
```diff
@@ -16,48 +17,89 @@
...
+
+
+@show_widget
+def test_suite_with_child_devices(suite, device):
     assert device in suite.devices
-    # Grab all device displays
-    assert len(suite._device_group.childs) == 1
-    child_displays = suite._device_group.childs[0].childs
+    device_group = suite.top_level_groups[0]
+    assert len(device_group.childs) == 1
```
klauer: Is it the right time to jettison the `childs` flotsam and rename to children?

typhon/suite.py
```diff
@@ -250,54 +219,60 @@ def hide_subdisplay(self, widget):
...
+        for group in self.top_level_groups:
+            for param in flatten_tree(group)[1:]:
+                self.hide_subdisplay(param)
 
     @property
     def tools(self):
         """Tools loaded into the DeviceDisplay"""
-        return [param.value() for param in self._tool_group.childs]
+        for group in self.top_level_groups:
+            if group.name() == 'Tools':
```
klauer: I think this may call for a `get_top_level_group_by_name('Tools')`

typhon/suite.py
```diff
@@ -326,6 +304,29 @@ def from_device(cls, device, parent=None,
...
             except Exception:
                 logger.exception("Unable to load %s", type(tool))
-        param = display.add_device(device, **kwargs)
-        display.show_subdisplay(param)
+        display.add_device(device, **kwargs)
+        display.show_subdisplay(device)
         return display
+
+    def _add_to_sidebar(self, parameter, category=None):
+        """Add an item to the sidebar, connecting neccesary signals"""
```
klauer:
```suggestion
        """Add an item to the sidebar, connecting necessary signals"""
```

typhon/suite.py
```diff
@@ -326,6 +304,29 @@ def from_device(cls, device, parent=None,
...
+        if category:
+            # Create or grab our category
+            group_dict = dict((param.name(), param)
+                              for param in self.top_level_groups)
```
klauer: Seeing this, maybe the above suggestion should be `get_top_level_group_dict()['Tools']` or something along those lines... 🚲 🏠

tests/test_suite.py
teddyrendahl: Unfortunately this part of the `pyqtgraph` API not mine 😢

typhon/suite.py
teddyrendahl: Maybe `top_level_groups` just returns the dictionary. You can always strip it down to `values()` for the same affect.

PR #146, opened 2018-12-14
Panel and TyphonDisplay Stylesheets

### LINKED ISSUE #141: Stylesheet customization of Panel and Display
You should be able to customize which templates are shown with which classes using a standard `qss` stylesheet. Same goes for individual signal widgets inside a `TyphonPanel`.

teddyrendahl:
## Description

Add a few utility properties and names for use with stylesheets:

### TyphonDisplay
Added read-only `pyqtproperty`s `device_name` and `device_class`. This allows you to set a template from a stylesheet like:
```css
TyphonDisplay [device_class='ophyd.sim.SynAxis'] {qproperty-force-template: '/path/to/my/template.ui'}
TyphonDisplay [device_name='test_motor']{...}
TyphonDisplay [device_name='test_motor'][display_type='1']{...}
```
##### Gotchas
* You have to use the integer spec for the `display_type` enum. This could be remedied by adding another property that is the string but I hate the redundancy and an obvious name for the new property escaped me

* The `device_class` is just string matching. No inheritance is respected. Probably not a huge deal but if we start having classes like `AreaDetectorV26` it might get obnoxious 😄 

* If you `force_template` for one of the display types you then need to specify them all. An empty string will use the default (either `happi` or `typhon`). This is because once you set a stylesheet that property is set until overwritten

### TyphonPanel
The `objectName` are set for all of the children widgets inside the layout based on the signal name. This is a nice way to get around the problem where we want units shown for some signals, not for others e.t.c. 

```css
PyDMLineEdit#my_motor_setpoint_edit {qproperty-showUnits: True}
```
## Motivation and Context

Closes #141 

## How Has This Been Tested?

Tests for new properties

### DIFF
--- a/typhon/display.py
+++ b/typhon/display.py
@@ -156,7 +156,7 @@ class TyphonDisplay(TyphonBase, TyphonDesignerMixin, DisplayTypes):
     def force_template(self, value):
         if value != self._forced_template:
             self._forced_template = value
-            self.load_template()
+            self.load_template(macros=self._last_macros)
 
     def add_device(self, device, macros=None):
         """
--- a/typhon/display.py
+++ b/typhon/display.py
@@ -146,6 +146,10 @@ class TyphonDisplay(TyphonBase, TyphonDesignerMixin, DisplayTypes):
             self._main_widget = QWidget()
         finally:
             self.layout().addWidget(self._main_widget)
+            # The following code reapplies the stylesheet
+            self.style().unpolish(self)
+            self.style().polish(self)
+            self.update()
 
     @Property(str)
     def force_template(self):
--- a/typhon/display.py
+++ b/typhon/display.py
@@ -108,6 +108,13 @@ class TyphonDisplay(TyphonBase, TyphonDesignerMixin, DisplayTypes):
                              device_class.__name__))
         return ''
 
+    @Property(str, designable=False)
+    def device_name(self):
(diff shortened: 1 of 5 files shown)


typhon/display.py
```diff
@@ -99,6 +99,22 @@ def display_type(self, value):
...
             self.load_template(macros=self._last_macros)
 
+    @Property(str, designable=False)
+    def device_class(self):
+        """Full class with module name of loaded device"""
+        if self.devices:
+            device_class = self.devices[0].__class__
+            return '.'.join((device_class.__module__,
+                             device_class.__name__))
+        return ''
```
ZLLentz: Is it possible to do something like:
```
@Property(type, designable=False):
    if self.devices:
        return self.devices[0].__class__
    return object
```
To send literally the class? Then it would be easy to do `isinstance` and `issubclass` as needed, rather than just matching strings. Though I wouldn't be surprised if `pyqt` prohibits this.

teddyrendahl: I could do that but how do you then compare that in a stylesheet?

ZLLentz: Just make stylesheets `.py` files

... just kidding. But maybe you have to do some import and `isinstance` magic either way, independent of swapping this.

klauer: There has to be a way... hmm...

PR #148, opened 2018-12-19
Method Button

### LINKED ISSUE #103: raise_to_operator
## Expected Behavior

Wrapper for a function. If we raise an exception during a call catch and raise in an error dialog

### LINKED ISSUE #126: Pushbutton Function Display
Simple `QPushButton` that runs a function, available in `Designer`. Option to respect a `Status` object. This could:
* Change stylesheet based on progress / success
* Disable until function is complete

teddyrendahl:
## Description

A button you can drag into `QtDesigner` to run a function. The button accepts an object (either added programmatically or via a `happi` plugin), a `method_name` and whether or not we `use_status` to give the button some state. By default, if `use_status` is `True` we disable the button while the underlying operation is in progress.

In order to show any error messages that may pop up during the execution of the function I made the general `raise_to_operator` method that displays an `Exception` and corresponding `traceback` in an `QMessageBox`

## Motivation and Context

Closes #103 
Closes #126 

## How Has This Been Tested?

Lots of tests. Coverage is going to be low on the `TyphonStatusThread` because it is run mostly inside a `QThread` however all the endpoints are checked in the test suite so it is surely being executed.

button

### DIFF
--- a/tests/test_func.py
+++ b/tests/test_func.py
@@ -56,9 +56,8 @@ class MyDevice(Device):
 @pytest.fixture(scope='function')
 def method_button(qtbot):
     dev = MyDevice(name='test')
-    button = TyphonMethodButton()
+    button = TyphonMethodButton.from_device(dev)
     qtbot.addWidget(button)
-    button.add_device(dev)
     button.method_name = 'my_method'
     return button
 
--- a/tests/test_func.py
+++ b/tests/test_func.py
@@ -1,17 +1,22 @@
 ############
 # Standard #
 ############
+import time
+import threading
 
 ############
 # External #
 ############
 import pytest
+from ophyd import Device
+from ophyd.status import Status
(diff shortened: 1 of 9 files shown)


ZLLentz:
Are you planning to expand this to passing arguments? I could see good points for and against this. It's hard to define these well from designer, but maybe it would be useful. It might also be neat to be able to link line edits, etc. into the button to set arguments live.

Anyway, LGTM.

teddyrendahl:
@ZLLentz Yeah, the `QPushButton` was so clear that is why I decided to implement it now. A large function is more difficult. I thought about just doing a single `QLineEdit` but it still wasn't quite clear if you want the function run when  return is pressed or when you are done typing, type conversion e.t.c. 

I kind of like the idea of being able drag arbitrary widgets in, label them as parameters with a type and then link them to the push button somehow.

PR #149, opened 2018-12-20
Typhon Command Line Utility

### LINKED ISSUE #131: Typhon Command Line Utility
- [ ] Load a device or multiple from `happi`
- [ ] Create a `TyphonSuite`

teddyrendahl:
## Description

Typhon command line utility to create a `TyphonSuite` with a group of devices. User passes in a list of devices by name and optionally:
* `--version` just to print information about where the utility comes from
* `--debug` to blast all information from the root logger. Otherwise the `INFO` stream is used from the `typhon` logger
* `--dark` to use `QDarkStyleSheet`
* `--stylesheet` if you want to pass in a custom stylesheet. In my mind this will be for specifying device specific screens e.t.c, not just base widget styling, but the option is there as well.
* `--happi-cfg` if you want to point at a specify `happi` configuration file. Otherwise environment variables are used.

Also saw a few more issues when testing on my fork with freezes on Travis. Explicitly disconnecting a few of my fixtures seemed to help but unsure if this was just a race condition I got lucky on a few times.

## Motivation and Context

Closes #131 

## How Has This Been Tested?

Tests added. By the way this way was way better than just putting a shell script in `bin`. Very simple to test since you can just import it the main function. Just have to be slightly careful about where you actually call the `parse_args`. Certainly going to continue to do it this way in the future.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,4 +18,4 @@ setup(name='typhon',
       include_package_data=True,
       install_requires=requirements,
       description='Interface generation for ophyd devices',
-      )
+      entry_points= {'console_scripts': ['typhon=typhon.cli:main']})
--- a/setup.py
+++ b/setup.py
@@ -18,4 +18,4 @@ setup(name='typhon',
       include_package_data=True,
       install_requires=requirements,
       description='Interface generation for ophyd devices',
-      entry_points= {'console_scripts': ['typhon=typhon.cli:main']})
+      entry_points={'console_scripts': ['typhon=typhon.cli:main']})
--- a/tests/test_display.py
+++ b/tests/test_display.py
@@ -1,5 +1,6 @@
 import os.path
 
+from pydm.utilities import close_widget_connections
 import pytest
 
 from typhon import TyphonDisplay
@@ -12,7 +13,8 @@ from .conftest import show_widget
 def display(qtbot):
     display = TyphonDisplay()
     qtbot.addWidget(display)
-    return display
+    yield display
+    close_widget_connections(display)
 
 
 @show_widget
--- a/typhon/cli.py
+++ b/typhon/cli.py
@@ -3,7 +3,6 @@ import logging
 import sys
 
 import coloredlogs
-import happi
 from qtpy.QtWidgets import QApplication
 
 import typhon
@@ -48,6 +47,12 @@ def typhon_cli(args):
     # Version endpoint
     if args.version:
         print(f'Typhon: Version {typhon.__version__} from {typhon.__file__}')
+        return
+
+    try:
+        import happi
+    except (ImportError, ModuleNotFoundError):
+        logger.exception("Unable to import happi to load devices!")
         return
 
     logger.debug("Creating Happi Client ...")
--- a/typhon/cli.py
+++ b/typhon/cli.py
@@ -1,3 +1,4 @@
+"""This module defines the ``typhon`` command line utility"""
 import argparse
 import logging
 import sys
@@ -28,6 +29,10 @@ parser.add_argument('--dark', action='store_true',
                     help='Use the QDarkStyleSheet shipped with typhon')
 parser.add_argument('--stylesheet',
                     help='Additional stylesheet options')
+
+
+# Append to module docs
+__doc__ += '\n::\n\n    ' + parser.format_help().replace('\n', '\n    ')
 
 
 def typhon_cli(args):
--- a/typhon/cli.py
+++ b/typhon/cli.py
@@ -20,7 +20,7 @@ parser.add_argument('devices', nargs='*',
 parser.add_argument('--happi-cfg',
                     help='Location of happi configuration file '
                          'if not specified by $HAPPI_CFG environment variable')
-parser.add_argument('--version', action='store_true',
+parser.add_argument('--version', '-V', action='store_true',
                     help='Current version and location '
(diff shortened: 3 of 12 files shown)


typhon/cli.py
```diff
@@ -0,0 +1,91 @@
...
+parser.add_argument('--version', action='store_true',
+                    help='Current version and location '
+                         'of Typhon installation.')
+parser.add_argument('--debug', action='store_true',
+                    help='Show the debug logging stream')
+parser.add_argument('--dark', action='store_true',
+                    help='Use the QDarkStyleSheet shipped with typhon')
+parser.add_argument('--stylesheet',
+                    help='Additional stylesheet options')
+
```
ZLLentz: One more cool thing you can do since you set up the cli like this is to include the `--help` text inside your sphinx docs, see https://github.com/pcdshub/hutch-python/blob/cde63c1b0c03adebe32741a2afb5fcba0557a83c/hutch_python/cli.py#L41 where my autodocs are picking up the module's `__doc__` string.

setup.py
```diff
@@ -18,4 +18,4 @@
       include_package_data=True,
       install_requires=requirements,
       description='Interface generation for ophyd devices',
-      )
+      entry_points= {'console_scripts': ['typhon=typhon.cli:main']})
```
ZLLentz: :+1:

typhon/cli.py
```diff
@@ -0,0 +1,91 @@
...
+            device = client.load_device(name=device)
+            suite.add_device(device)
+            suite.show_subdisplay(device)
+        except Exception:
+            logger.exception("Unable to add %r to TyphonSuite", device)
+
+    # Deal with stylesheet
+    typhon.use_stylesheet(dark=args.dark)
+    if args.stylesheet:
+        logger.info("Loading QSS file %r ...", args.stylesheet)
```
ZLLentz: What's qss?

typhon/cli.py
```diff
@@ -0,0 +1,91 @@
...
+        with open(args.stylesheet, 'r') as handle:
+            app.setStyleSheet(handle.read())
+
+    logger.info("Launching application ...")
+    suite.show()
+    app.exec_()
+    logger.info("Execution complete!")
```
ZLLentz: Such a cheery exit message

typhon/cli.py
teddyrendahl: This is a very good idea

typhon/cli.py
teddyrendahl: **Q**t **S**tyle**s**heet

ZLLentz: Gotcha, thanks. It wasn't on the first page of Google results for `qss`

typhon/cli.py
```diff
@@ -0,0 +1,96 @@
...
+parser = argparse.ArgumentParser(description='Create a TyphonSuite for '
+                                             'device/s stored in a Happi '
+                                             'Database')
+
+parser.add_argument('devices', nargs='*',
+                    help='Device names to load in the TyphonSuite')
+parser.add_argument('--happi-cfg',
+                    help='Location of happi configuration file '
+                         'if not specified by $HAPPI_CFG environment variable')
+parser.add_argument('--version', action='store_true',
```
klauer: Having `-V` as an alias for this would also be nice

typhon/cli.py
```diff
@@ -0,0 +1,96 @@
...
+                    help='Current version and location '
+                         'of Typhon installation.')
+parser.add_argument('--debug', action='store_true',
```
klauer: Having `-v` for verbosity would be slightly more standard

typhon/cli.py
```diff
@@ -0,0 +1,96 @@
...
+            suite.show_subdisplay(device)
+        except Exception:
+            logger.exception("Unable to add %r to TyphonSuite", device)
+
+    # Deal with stylesheet
+    typhon.use_stylesheet(dark=args.dark)
+    if args.stylesheet:
+        logger.info("Loading QSS file %r ...", args.stylesheet)
+        with open(args.stylesheet, 'r') as handle:
+            app.setStyleSheet(handle.read())
```
klauer: It'd be nice if the stylesheet was verified to exist prior to creating any widgets, in the case of a mistyped filename

tests/test_display.py
```diff
@@ -12,7 +13,8 @@
 def display(qtbot):
     display = TyphonDisplay()
     qtbot.addWidget(display)
-    return display
+    yield display
```
klauer: pytest fixtures can be used as context managers without even a decorator? Neat!

setup.py
klauer:
```suggestion
      entry_points={'console_scripts': ['typhon=typhon.cli:main']})
```

tests/test_display.py
teddyrendahl: https://docs.pytest.org/en/latest/yieldfixture.html 😄

typhon/cli.py
teddyrendahl: ✅ Though not going to bother with multiple verbosity flags i.e `-vvv`

teddyrendahl:
Power cycling to re-run tests

teddyrendahl:
Struggling with the test suite. Merging this then going to follow up with another PR that fixes things. This code is not the issue.

PR #150, opened 2018-12-21
No happi no cry

teddyrendahl:
## Description

Make `happi` an optional dependency. The only place it was imported before without a `try/except` is in the `HappiPlugin`. This is now protected. In the future we can make a compatibility layer with whatever other device server we happen upon. But until we find a second alternative to `happi` I'm going to hold off.

## Motivation and Context

@tacaswell 

## How Has This Been Tested?

Added a test, largely to protect against regression


tacaswell:
:+1: Thank you!

PR #152, opened 2019-01-08
Plugin and Test Suite Cleanup

teddyrendahl:
## Description

The test suite is a little unstable and I think may have finally figured out why. In `PyDM v1.6.0` there is now a `destroying` keyword on disconnect which basically lets us know that the `QWidget` which holds the `PyDMChannel` is being garbage collected, not just disconnected. So instead of making sure everything is disconnected on an object that is "between worlds", we just let it pass by. The fundamental reason for this a lot of 💩happens when an `QObject` still has a `Python` representation but doesn't exist in the `C++` layer or vice versa. The `destroying` key was not implemented in either of the `Typhon` plugins which I think is why we were seeing segfaults on fixtures that were continually torn down and re-initialized. 

Secondly, with the `status` tests I was seeing some hangs when running on Travis. This shouldn't really be possible with the `qtbot.waitSignal`, but I have a feeling that the context manager wasn't quite working like it should. I also made sure that any threads started by the fixture are ended before we start our next test.

### DIFF
--- a/typhon/plugins/core.py
+++ b/typhon/plugins/core.py
@@ -157,7 +157,7 @@ class SignalConnection(PyDMConnection):
                     logger.debug("%s has no value_signal for type %s",
                                  channel.address, _typ)
 
-    def remove_listener(self, channel, **kwargs):
+    def remove_listener(self, channel, destroying=False, **kwargs):
         """
         Remove a listener channel from this connection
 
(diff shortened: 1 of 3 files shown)


typhon/plugins/core.py
```diff
@@ -87,8 +88,9 @@ def put_value(self, new_val):
                 new_val = self.signal_type(new_val)
             logger.debug("Putting value %r to %r", new_val, self.address)
             self.signal.put(new_val)
-        except Exception:
+        except Exception as exc:
             logger.exception("Unable to put %r to %s", new_val, self.address)
+            raise_to_operator(exc)
```
klauer: 👍

hhslepicka: 👍

hhslepicka:
You are absolutely right. This "between worlds" state was a surprise for me too.
Sorry for the bad job announcing this finding and breaking the API.

PR #155, opened 2019-01-11
API Breaks and Deprecations

teddyrendahl:
Last round of API breaks I wanted to get in before I tag:

## API
I really didn't like how vague the names were for `TyphonPanel` and `TyphonDisplay`. They convey very little meaning. They are now `TyphonDeviceDisplay` and `TyphonSignalPanel`. A little wordier but easier to tell what they are going to do

## Deprecations
After a conversation with @klauer, lugging a bunch of API from `v0.1.0` around probably isn't necessary. Now that every `typhon` widget has `.from_device` class constructor I don't need factory functions so that removes any need for `DeviceDisplay` and `DeviceTimePlot`. There was also an old utility called `grab_hints` which has been replaced by the more general `grab_kind`

Also a few small fixes:
*  The `typhon` command line function was not loading any of the tools
*  The `typhon` widgets in the Designer were under the "PyDM Widgets" group

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ by the operator:
 * ``TyphonSuite``: The overall view for a Typhon window. It allows the
 operator to view all of the loaded components and tools
 
-* ``TyphonDisplay``: This is the widget created for a standard
+* ``TyphonDeviceDisplay``: This is the widget created for a standard
 ``ophyd.Device``. Signals are organized based on their
 ``Kind`` and description.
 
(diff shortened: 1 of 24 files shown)


klauer:
👍 on the renames and cleaning up deprecations. PR message incomplete?
 
> There was also an old utility called grab_hints which has been replaced by the more general `

teddyrendahl:
Yup, got distracted. Thanks @klauer

klauer:
Verifying if anything remains of the old names:
```bash
$ git grep TyphonDisplay
README.md:* ``TyphonDisplay``: This is the widget created for a standard
$ git grep TyphonPanel
$ git grep grab_hints
$
```

PR #163, opened 2019-02-05
Suite saving

### LINKED ISSUE #17: Save generated user interfaces

teddyrendahl:
## Description

Woo hoo! Closing the 4th oldest issue in the library.

The idea here is if I add a bunch of devices to my `TyphonSuite` then decide I will want to reload this later, how can I save and reload this object. After a conversation with @klauer and @hhslepicka I decided to see how the route of actually saving the Python code required to generate the ``TyphonSuite`` looked. A button was added to the ParameterTree so that this is available via the GUI.

I did something which I think is clever here and broke up the `typhon.cli` library in order make a terse template that either can be executed as a standalone Python script or imported and used as part of a bigger Qt application. By using the `typhon.cli` you will still have access to all the additional options:

```
$ python my_saved_suite.py --verbose
```

But I can load as many `TyphonSuite` files as I want using `typhon.load_suite`

## Motivation and Context

Closes #17 

## How Has This Been Tested?

Tests added for all new utilities

Narrative doc page added with explanation, API docs as well.

screen shot 2019-02-04 at 8 59 33 pm

### DIFF
--- a/docs/source/save.rst
+++ b/docs/source/save.rst
@@ -24,7 +24,7 @@ to disk.
 There are two major ways to use this created file:
 
 1. Execute the Python file from the command line. This will route the call
-   through the standard :func:`.typhon.cli.typhon_cli` meaning all options
+   through the standard :func:`.cli.typhon_cli` meaning all options
    described there are also available.
 
 .. code:: bash
--- a/docs/source/save.rst
+++ b/docs/source/save.rst
@@ -3,28 +3,15 @@ Saving and Loading
 ##################
 :class:`.TyphonSuite` objects can be stored for later use. The devices that
 were loaded into the suite via :meth:`.TyphonSuite.add_device` will be added
-once again assuming that they are stored in a ``happi`` database. When you call
-:meth:`.TyphonSuite.save` a Python file with the template shown below is saved
-to disk.
+once again assuming that they are stored in a ``happi`` database.
 
-.. code:: python
-
-    import sys
-    import typhon.cli
-
-    devices = {devices}
-
-    def create_suite(cfg=None):
-        return typhon.cli.create_suite(devices, cfg=cfg)
-
-    if __name__ == '__main__':
-        typhon.cli.typhon_cli(devices + sys.argv[1:])
-
+.. automethod:: typhon.TyphonSuite.save
+   :noindex:
 
 There are two major ways to use this created file:
 
 1. Execute the Python file from the command line. This will route the call
-   through the standard :func:`.cli.typhon_cli` meaning all options
+   through the standard :mod:`typhon.cli` meaning all options
    described there are also available.
 
 .. code:: bash
--- a/docs/source/save.rst
+++ b/docs/source/save.rst
@@ -1,0 +1,64 @@
+##################
+Saving and Loading
+##################
+:class:`.TyphonSuite` objects can be stored for later use. The devices that
+were loaded into the suite via :meth:`.TyphonSuite.add_device` will be added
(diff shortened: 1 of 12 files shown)


docs/source/save.rst
```diff
@@ -0,0 +1,64 @@
...
+
+    import sys
+    import typhon.cli
+
+    devices = {devices}
+
+    def create_suite(cfg=None):
+        return typhon.cli.create_suite(devices, cfg=cfg)
+
+        if __name__ == '__main__':
```
klauer: Indentation?

docs/source/save.rst
```diff
@@ -0,0 +1,64 @@
...
+            typhon.cli.typhon_cli(devices + sys.argv[1:])
+
+
+There are two major ways to use this created file:
+
+1. Execute the Python file from the command line. This will route the call
+   through the standard :func:`.typhon.cli.typhon_cli` meaning all options
```
klauer:
```suggestion
   through the standard :func:`.cli.typhon_cli` meaning all options
```
(Maybe? Seems inconsistent with the other sphinx directives...)

tests/test_utils.py
```diff
@@ -74,3 +77,21 @@ def test_raise_to_operator_msg(monkeypatch, qtbot):
     qtbot.addWidget(exc_dialog)
     assert exc_dialog is not None
     assert 'ZeroDivisionError' in exc_dialog.text()
+
+
+def test_load_suite(qtbot):
+    # Setup new saved file
+    module = saved_template.format(devices=[])
+    module_file = str(pathlib.Path(tempfile.gettempdir()) / 'my_suite.py')
```
klauer: Temp file not deleted?

typhon/suite.py
```diff
@@ -355,6 +360,27 @@ def from_device(cls, device, parent=None,
...
+        logger.debug("Requesting file location for saved TyphonSuite")
+        root_dir = str(pathlib.Path(__file__).parent)
+        filename = QFileDialog.getSaveFileName(self, 'Save TyphonSuite',
+                                               root_dir, "Python (*.py)")
+        if filename:
+            try:
+                with open(filename[0], 'w+') as handle:
+                    save_suite(self, handle)
+            except Exception as exc:
+                logger.error("Failed to save TyphonSuite")
```
klauer:
```suggestion
                logger.exception("Failed to save TyphonSuite")
```

typhon/cli.py
```diff
@@ -54,48 +54,65 @@ def typhon_cli(args):
...
-        logger.exception("Unable to import happi to load devices!")
-        return
-
-    logger.debug("Creating Happi Client ...")
-    client = happi.Client.from_config(cfg=args.happi_cfg)
-
-    logger.debug("Creating widgets ...")
-    app = QApplication.instance()
+    # Deal with stylesheet
+    logger.debug("Applying stylesheet ...")
```
ZLLentz: This debug message seems misplaced, since the next thing done here is create the app rather than call `use_stylesheet`

docs/source/save.rst
klauer: Seems reasonable to shove it in a docstring and display it that way.

PR #165, opened 2019-02-07
[WIP] CI: azure pipelines

klauer:
(Based on https://github.com/hhslepicka/pydm/tree/azure-pipelines)

3.6/3.7 tagged versions pass.
I think if we had deps for Windows/OSX, they might pass as well.

Remaining:
* Deploying packages
* Documentation
* Remove and/or allow failure of Win/OSX

### DIFF
--- a/.azure/template.yml
+++ b/.azure/template.yml
@@ -1,0 +1,108 @@
+parameters:
+  name: ''
+  vmImage: ''
+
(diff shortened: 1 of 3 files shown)


klauer:
Cycling to nudge azure (hopefully enabled now...)

PR #168, opened 2019-02-08
Array Data

### LINKED ISSUE #97: Handle Array Data
Currently `signal_widget` contains routing for all `dtype` signals with the exception of arrays. These should either be sent to a PyDM plot or image view based on the `shape`

teddyrendahl:
## Description

Previously `typhon` just threw array data on the floor when it found it in a `Signal`. With this PR we now create a `QPushButton` that either launches a `PyDMWaveformPlot` for a one-dimensional array or a `PyDMImageView` for two-dimensional. Unfortunately, PyDM doesn't deal with colored images so still can't handle that.

I think this is better for the long term instead of just showing them directly inside the grid. It doesn't scale well for multiple arrays, but I'm open to alternative opinions. My thought is you can always customize your template to include a full size widget.

I also think that using my test suite to visually check all the widgets might have reached a tipping point. I put a small examples folder in with a fake device and panel to mess around with.

## Motivation and Context

Closes #97 

## How Has This Been Tested?

Tests for new widgets and in context of `signal_widget` creation

New `examples` added

panel_with_arrays

### DIFF
--- a/examples/panel.py
+++ b/examples/panel.py
@@ -12,7 +12,7 @@ class Sample(Device):
     readback = Cpt(SignalRO, value=1)
     setpoint = Cpt(Signal, value=2)
     waveform = Cpt(SignalRO, value=np.random.randn(100, ))
-    image = Cpt(SignalRO, value=np.random.randn((100, 100)) * 455)
+    image = Cpt(SignalRO, value=np.abs(np.random.randn(100, 100)) * 455)
 
 
 # Create my device without a prefix
--- a/examples/panel.py
+++ b/examples/panel.py
@@ -12,7 +12,7 @@ class Sample(Device):
     readback = Cpt(SignalRO, value=1)
     setpoint = Cpt(Signal, value=2)
     waveform = Cpt(SignalRO, value=np.random.randn(100, ))
-    image = Cpt(SignalRO, value=np.ones((100, 100)) * 455)
+    image = Cpt(SignalRO, value=np.random.randn((100, 100)) * 455)
 
 
 # Create my device without a prefix
--- a/examples/panel.py
+++ b/examples/panel.py
@@ -1,0 +1,30 @@
+"""Example to create a Panel of Ophyd Signals from an object"""
+import sys
+import numpy as np
+from ophyd import Device, Component as Cpt, Signal
+from ophyd.sim import SignalRO
+from qtpy.QtWidgets import QApplication
+import typhon
+
+
+class Sample(Device):
+    """Simulated Device"""
+    readback = Cpt(SignalRO, value=1)
+    setpoint = Cpt(Signal, value=2)
+    waveform = Cpt(SignalRO, value=np.random.randn(100, ))
+    image = Cpt(SignalRO, value=np.ones((100, 100)) * 455)
+
+
+# Create my device without a prefix
+sample = Sample('', name='sample')
+
+if __name__ == '__main__':
+    # Create my application
+    app = QApplication(sys.argv)
+    typhon.use_stylesheet()
+    # Create my panel
+    panel = typhon.TyphonSignalPanel.from_device(sample)
+    panel.sortBy = panel.byName
+    # Execute
+    panel.show()
+    app.exec_()
--- a/typhon/plugins/core.py
+++ b/typhon/plugins/core.py
@@ -70,6 +70,33 @@ class SignalConnection(PyDMConnection):
         # Add listener
         self.add_listener(channel)
 
+    def cast(self, value):
+        """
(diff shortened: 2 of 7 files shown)


examples/panel.py
```diff
@@ -0,0 +1,30 @@
...
+from qtpy.QtWidgets import QApplication
+import typhon
+
+
+class Sample(Device):
+    """Simulated Device"""
+    readback = Cpt(SignalRO, value=1)
+    setpoint = Cpt(Signal, value=2)
+    waveform = Cpt(SignalRO, value=np.random.randn(100, ))
+    image = Cpt(SignalRO, value=np.ones((100, 100)) * 455)
```
klauer:
```suggestion
    image = Cpt(SignalRO, value=np.random.randn((100, 100)) * 455)
```
This would make for a more interesting image, at least.

typhon/widgets.py
```diff
@@ -238,3 +239,63 @@ def channel(self, value):
     def _tx(self, value):
         """Receive information from happi channel"""
         self.add_device(value['obj'])
+
+
+class SignalDialogButton(QPushButton):
+    """QPushButton to launch a QDialog with a PyDMWidget"""
+    text = NotImplemented
```
klauer: Should these be customizable at `__init__` time as well?

typhon/plugins/core.py
```diff
@@ -70,6 +70,33 @@ def __init__(self, channel, address, protocol=None, parent=None):
...
+        through the correct signal and convert values returned by the widget to
+        the correct type before handing them to Ophyd Signal
+        """
+        # If this is the first time we are receiving a new value note the type
+        # We make the assumption that signals do not change types during a
+        # connection
+        if not self.signal_type:
+            dtype = self.signal.describe()[self.signal.name]['dtype']
+            # Only way this raises a KeyError is if ophyd is confused
+            self.signal_type = _type_map[dtype][0]
```
klauer: I wonder if in ophyd there should be a suggested type to cast to. Picking the first one works in all cases here, but feels a bit off.

typhon/signal.py
```diff
@@ -56,29 +57,44 @@ def signal_widget(signal, read_only=False):
...
-        # Otherwise a LineEdit will suffice
+    # Grab a description of the widget to see the correct widget type
+    try:
+        desc = signal.describe()[signal.name]
+    except Exception:
+        logger.error("Unable to connect to %r during widget creation",
+                     signal.name)
+        desc = {}
+    # Unshaped data
+    if desc.get('shape', []) == []:
```
klauer:
```suggestion
    shape = desc.get('shape', [])
    try:
        dimensions = len(shape)
    except TypeError:
        dimensions = 0
    if dimensions == 0:
```

typhon/signal.py
```diff
@@ -56,29 +57,44 @@ def signal_widget(signal, read_only=False):
...
+            # Otherwise a LineEdit will suffice
+            else:
+                logger.debug("Creating LineEdit for %s", signal.name)
+                widget = TyphonLineEdit
+                name = signal.name + '_edit'
+    # Waveform
+    elif len(desc.get('shape')) == 1:
+        logger.debug("Creating WaveformDialogButton for %s", signal.name)
+        widget = WaveformDialogButton
+        name = signal.name + '_waveform_button'
```
klauer: It's probably unnecessary to do this any more dynamically, but perhaps at least you could make a `{widget: suffix}` dictionary and create the signal name below this big if block? All these `debug` messages could also be wrapped into one message as `widget` and `name` give you the full string.

typhon/plugins/core.py
teddyrendahl: Agreed. I'm all for more being available via introspection of `Signal`. At this point I think this is the best I can do.

typhon/signal.py
teddyrendahl: In case `shape` has no `len`?

klauer: It really _shouldn't_ be the case (as it would break the data model), but you also wouldn't want it crashing here unnecessarily in case `None` or a scalar gets through.

typhon/widgets.py
teddyrendahl: Yeah I can add these. It did bother me that the base class was unusable.

typhon/signal.py
teddyrendahl: I want to refactor this whole function. I think just a single row of the panel should be usable in designer. I also want to make sure that you can indicate via stylesheet if you want a specific type of widget there. I.e Show `low_limit_switch` as an indicator light.

PR #170, opened 2019-02-27
ENH: AreaDetector port flow chart user interface using pyqtgraph

klauer:
## Motivation and Context
Visualizing port connectivity in AreaDetector is a pain. ophyd supplies a networkx-based solution for a simple, static view of the ports, but it does not allow for real-time viewing or GUI-based reconfiguration. This PR aims to make port configuration easy through an interactive graph visualization of the nodes.

* All that is necessary is a correctly-configured ophyd Detector object with all of the plugins 
* It automatically updates as the port configuration is changed on the EPICS level: both the hierarchical tree and the node graph
* Click-and-drag to configure a plugin chain (must click the 'Out' terminals and 'In' terminals)
* Drag-and-drop on tree reconfigures ports and the node graph

TODO
* Make automatic node-repositioning optional such that the user can move around nodes as they please
* Add the ability to live-preview images (or maybe even files, if accessible from the client machine? ... Hmm, probably not...)
* Add the ability to configure a plugin with the typhon-generated screens after clicking on one

## Screenshots
(Outdated screenshot with overlap issue)
port_demo2

## How Has This Been Tested?
Interactively, and with the included unit tests.

It hasn't.

## Pending
* A few ophyd PRs are necessary for this to work (AD plugin one, new callback issue, ...)
* See the pyqtgraph section below

## pyqtgraph.Flowchart
pyqtgraph's Flowchart hasn't been updated for around a year and still contains some Python 2-only stuff (`unicode` in various places?). The classes weren't particularly built with subclassing in mind, and is spaghetti-like in the widget relationships - resulting in some of the mess that's here in this PR. 

There may be a good case to either:
1. Vendor the flowchart code from pyqtgraph, updating and cleaning it
2. Find an alternative library to base this on

For (2), this PR provides a couple of classes and utility functions that will be applicable regardless of the widgets used to display the graph.


klauer:
As discussed, since this is beyond the scope of typhon, this will move to a high-level application repository of its own.  (`typhon-adviewer`?)

PR #173, opened 2019-03-06
REF: proof-of-concept for hiding Q_ENUMS a bit better

klauer:
_Concept:_
This is the best I've been able to come up with due to how PyQt implements `Q_ENUMS` under the hood. **This doesn't have to be merged.**

The return value of the `UsesEnums` helper is actually a base class, so that's why it's in camel case:

Usage:
```python
class TyphonSignalPanel(UsesEnums(SignalOrder), TyphonBase, TyphonDesignerMixin):
```

I thought about that a bit more, then found I liked the following better:
```python
QtSignalOrder = UsesEnums(SignalOrder)

class TyphonSignalPanel(QtSignalOrder, TyphonBase, TyphonDesignerMixin):
```

If the contents of this PR are really found to be desirable, I think just renaming to `make_qt_enum_base_class` or something along those lines would probably be the next step. Then making all other classes use it.

_Proof:_
image

Reference:
PyQt4: https://github.com/Werkov/PyQt4/blob/master/qpy/QtCore/qpycore_qmetaobject.cpp#L45
PyQt5: https://github.com/baoboa/pyqt5/blob/master/qpy/QtCore/qpycore_enums_flags.cpp

### DIFF
--- a/typhon/utils.py
+++ b/typhon/utils.py
@@ -34,7 +34,8 @@ ui_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'ui')
 
 def UsesEnums(*enum_classes, base=QWidget):
     '''
-    Add Qt Designer compatible enums to a class.
+    Add Qt Designer compatible enums to a class. Under the hood, this generates
+    a base class for the target class.
 
     Parameters
     ----------
@@ -46,17 +47,22 @@ def UsesEnums(*enum_classes, base=QWidget):
         QObject or QWidget (or have one in the class hierarchy)
     '''
     class EnumBase(base):
-        '''
-
-        Attributes
-        ----------
-        _enums : tuple
-            Tuple of Qt-compatible enum classes used
-        '''
         for enum in enum_classes:
             Q_ENUMS(enum)
 
-        _enums = enum_classes
+        _qt_enums = enum_classes
+
+        @classmethod
+        def _find_qt_enums(class_):
+            'Return the set of all used Q_ENUMS on this class'
+            return set(
+                sum((cls._qt_enums
+                     for cls in class_.mro()
+                     if isinstance(getattr(cls, '_qt_enums', None), tuple)
+                     ),
+                    tuple()
+                    )
+            )
 
     for enum in enum_classes:
         # Allow access via cls.enum_class_name.member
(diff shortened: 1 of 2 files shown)


typhon/utils.py
```diff
@@ -31,6 +32,57 @@
...
+        QObject or QWidget (or have one in the class hierarchy)
+    '''
+    class EnumBase(base):
+        for enum in enum_classes:
+            Q_ENUMS(enum)
+
+        _qt_enums = enum_classes
+
+        @classmethod
+        def _find_qt_enums(class_):
```
teddyrendahl: This doesn't get called anywhere? What is its purpose?

teddyrendahl:
Ooh I like this. Think current Travis failure looks like some sort of API change on PyDM. I'll look into that. 

I think we should adopt this. We can discuss API but I kind of like `UsesEnums` better than your suggestion in the PR description.

typhon/utils.py
klauer: To `Return the set of all used Q_ENUMS on this class`
It just adds extra introspection capabilities to anything using `EnumBase`. If not deemed useful, both `_qt_enums` and `_find_qt_enums` could be nuked.

teddyrendahl: I think maybe keep `_qt_enums` but drop the discovery function? Not sure what the advantage is of having both, but maybe you could explain in person.

typhon/utils.py
```diff
@@ -31,6 +32,57 @@
...
+        where one attribute is associated with one value.
+    base : class
+        The base class for the generated Qt-compatible class. Must be either
+        QObject or QWidget (or have one in the class hierarchy)
+    '''
+    class EnumBase(base):
+        for enum in enum_classes:
+            Q_ENUMS(enum)
+
+        _qt_enums = enum_classes
```
teddyrendahl: What happens here if I inherit from a class that also inherits from `UseEnums`? Are we going to overwrite one?

klauer: You answered your above question with this one. `_qt_enums` gets wiped away, hence the need for the discovery function above.

Thinking a bit more about it, I suppose this could be refactored with an `__init_subclass__`-style thing to keep the list up-to-date, though that would pin the discovery functionality to 3.6+ of course.

klauer:
👎 The enum namespace has `EnumBase` in the name. 

While getting around that is possible, it requires _way too much_ Python magic (and that's coming from me...).

klauer:
For the record, this is what it looks like... things get real when `exec` is involved:
https://github.com/klauer/typhon/commit/e5d402704c81fac36b1187ada327f119998cc89c

PR #175, opened 2019-03-07
Small bugs

teddyrendahl:
## Description

Found a few of these when setting up the demo and also included a patch that is required after https://github.com/slaclab/pydm/pull/470

1. If we fail to call a function it will not produce a `Status`. We now have a check to see that we are ready for this scenario.
2. Not all widgets inherit from `TyphonBase`, the `TyphonMethodButton` for instance. We still want to auto populate as much as we can, i.e if a user drags in a `TyphonMethodButton` into their template we should fill it.
3. The `_unit` variable no longer exists on `PyDMWidgetBase` but instead only on widgets that inherit from `TextFormatter`. Simply using ` PyDMLineEdit` in the test will work and be back compatible.

### DIFF
--- a/tests/plugins/test_core.py
+++ b/tests/plugins/test_core.py
@@ -18,19 +18,11 @@ from typhon.plugins.core import (SignalPlugin, SignalConnection,
                                  register_signal)
 
 
-
-class WritableWidget(PyDMLineEdit):
-    """Simple Testing Widget"""
-    pass
-
-
-
-
 def test_signal_connection(qapp, qtbot):
     # Create a signal and attach our listener
     sig = Signal(name='my_signal', value=1)
     register_signal(sig)
-    widget = WritableWidget()
+    widget = PyDMLineEdit()
     qtbot.addWidget(widget)
     widget.channel = 'sig://my_signal'
     listener = widget.channels()[0]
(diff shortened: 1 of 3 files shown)


tests/plugins/test_core.py
```diff
@@ -19,7 +19,7 @@
 
 
 
-class WritableWidget(QWidget, PyDMWritableWidget):
+class WritableWidget(PyDMLineEdit):
```
klauer: Maybe rename the derived class?

typhon/display.py
```diff
@@ -153,7 +153,11 @@ def load_template(self, macros=None):
                                         macros=macros)
             # Add device to all children widgets
             if self.devices:
-                for widget in self._main_widget.findChildren(TyphonBase):
+                designer = (self._main_widget.findChildren(TyphonDesignerMixin)
+                            or [])
+                bases = (self._main_widget.findChildren(TyphonBase)
+                         or [])
+                for widget in set(bases + designer):
```
klauer: This is OK assuming pyqt, as `findChildren` is defined to return a `list`. I'm not sure about pyside (or if we care about it), as it says: 
image

teddyrendahl: I don't think we are promising support for `PySide` users.

PR #185, opened 2019-03-14
CLI Cleanup

### LINKED ISSUE #184: Command Line do not load Suite if no devices loaded
If you ask for a specific `happi` object via the cli that does not exist, do not load an empty `TyphonSuite`

teddyrendahl:
Noticed then when actually trying out the `typhon.cli` in the wild. If you mistype your device name i.e: `typhon not_a_device`. The CLI goes through all the trouble of creating a `TyphonSuite` for you that contains **no** devices. Not optimal.

This PR rearranges the code in the CLI function so we exit as early as possible if you entered malformed device information. The code is a little messier but I think this is better for the end user. The other commits were centered around the fact that I created and loaded a device-less `TyphonSuite` to test the command line interface so those tests had to be updated.

Closes #184

### DIFF
--- a/typhon/cli.py
+++ b/typhon/cli.py
@@ -97,7 +97,8 @@ def create_suite(devices, cfg=None):
         logger.info("Loading Tools ...")
         for name, tool in suite.default_tools.items():
             suite.add_tool(name, tool())
-        logger.info("Adding devices ...")
+        if devices:
+            logger.info("Adding devices ...")
         for device in loaded_devs:
             try:
                 suite.add_device(device)
--- a/typhon/cli.py
+++ b/typhon/cli.py
@@ -91,7 +91,7 @@ def create_suite(devices, cfg=None):
             loaded_devs.append(device)
         except Exception:
             logger.exception("Unable to load %r", device)
-    if loaded_devs:
+    if loaded_devs or not devices:
         logger.debug("Creating empty TyphonSuite ...")
         suite = typhon.TyphonSuite()
         logger.info("Loading Tools ...")
--- a/typhon/cli.py
+++ b/typhon/cli.py
@@ -40,11 +40,10 @@ def typhon_cli_setup(args):
     global app
     # Logging Level handling
     logging.getLogger().addHandler(logging.NullHandler())
+    shown_logger = logging.getLogger('typhon')
     if args.verbose:
         level = "DEBUG"
-        shown_logger = logging.getLogger('typhon')
     else:
-        shown_logger = logging.getLogger()
         level = "INFO"
     coloredlogs.install(level=level, logger=shown_logger,
                         fmt='[%(asctime)s] - %(levelname)s -  %(message)s')
--- a/typhon/cli.py
+++ b/typhon/cli.py
@@ -39,6 +39,7 @@ __doc__ += '\n::\n\n    ' + parser.format_help().replace('\n', '\n    ')
 def typhon_cli_setup(args):
     global app
     # Logging Level handling
+    logging.getLogger().addHandler(logging.NullHandler())
     if args.verbose:
         level = "DEBUG"
         shown_logger = logging.getLogger('typhon')
--- a/typhon/cli.py
+++ b/typhon/cli.py
@@ -82,21 +82,30 @@ def create_suite(devices, cfg=None):
     else:
         logger.debug("Creating new happi Client from configuration")
         client = happi.Client.from_config(cfg=cfg)
-    logger.debug("Creating empty TyphonSuite ...")
-    suite = typhon.TyphonSuite()
-    logger.info("Loading Tools ...")
-    for name, tool in suite.default_tools.items():
-        suite.add_tool(name, tool())
     # Load and add each device
+    loaded_devs = list()
     for device in devices:
         logger.info("Loading %r ...", device)
         try:
             device = client.load_device(name=device)
-            suite.add_device(device)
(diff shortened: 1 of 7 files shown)


teddyrendahl:
Had to rebase so I could get the `happi/cli.py` file that is causing the `flake8` exception on Travis 🤕

typhon/cli.py
```diff
@@ -82,21 +82,30 @@ def create_suite(devices, cfg=None):
...
         logger.debug("Creating new happi Client from configuration")
         client = happi.Client.from_config(cfg=cfg)
-    logger.debug("Creating empty TyphonSuite ...")
-    suite = typhon.TyphonSuite()
-    logger.info("Loading Tools ...")
-    for name, tool in suite.default_tools.items():
-        suite.add_tool(name, tool())
     # Load and add each device
+    loaded_devs = list()
     for device in devices:
```
ZLLentz: Is it possible to enter this block with an empty devices list? If not, then this is a good change.

klauer: So long as `devices` is an empty list/iterable and not `None`, it seems like this should be OK.

It will, however, return `None` instead of a suite as there will be no `loaded_devs`.

teddyrendahl: `devices` is either entered by the user via `create_suite` or comes directly from `ArgumentParser` if executing via the command line. The latter is guaranteed to be iterable, the first is not, but I'm not sure I need to explicitly have an error message for this?

Also to @klauer point, the method that calls this watches for a `None` being returned. I didn't want to raise an exception here and then catch it somewhere else later. Whether this is the cleanest way to implement this is up for debate.

ZLLentz: There should definitely be a message about why nothing happened if it's possible for a user to get here. Otherwise it is fine as is.

teddyrendahl: Point heard. This silently does nothing now which is the worst of both worlds. Now my question is if the user just calls the cli like this:

`typhon`

Do we?
* Count this an error, yell loudly you failed to enter any devices.
* Give them back an empty `TyphonSuite`. This option maybe more appealing someday if you can drag and drop new devices in from another application?

klauer: Empty suite may be preferable?

teddyrendahl: Done. Tests added as behavior is slightly complicated.

typhon/cli.py
```diff
@@ -82,21 +82,30 @@ def create_suite(devices, cfg=None):
...
-            logger.exception("Unable to add %r to TyphonSuite", device)
-    return suite
+            logger.exception("Unable to load %r", device)
+    if loaded_devs or not devices:
+        logger.debug("Creating empty TyphonSuite ...")
+        suite = typhon.TyphonSuite()
+        logger.info("Loading Tools ...")
+        for name, tool in suite.default_tools.items():
+            suite.add_tool(name, tool())
+        logger.info("Adding devices ...")
```
ZLLentz:
```suggestion
        if devices:
            logger.info("Adding devices ...")
```

teddyrendahl:
This should launch the `TyphonSuite` in a `QMainWindow`.

teddyrendahl:
Also some change in this PR is making the `--verbose` mode have duplicate logging messages:
```
[2019-03-19 15:28:57] - DEBUG -  Adding 'Log' to category 'Tools' ...
[2019-03-19 15:28:57] - DEBUG -  Connecting parameter signals ...
[2019-03-19 15:28:58] - DEBUG -  Adding widget 'StripTool' with <typhon.tools.plot.TyphonTimePlot object at 0x11723c678> to 'Tools' ...
[2019-03-19 15:28:58] - DEBUG -  Adding 'StripTool' to category 'Tools' ...
[2019-03-19 15:28:58] - DEBUG -  Connecting parameter signals ...
[2019-03-19 15:28:58] - DEBUG -  Starting Jupyter Kernel ...
[2019-03-19 15:28:58] - DEBUG -  Adding widget 'Console' with <typhon.tools.console.TyphonConsole object at 0x1175bfee8> to 'Tools' ...
DEBUG:typhon.suite:Adding widget 'Console' with <typhon.tools.console.TyphonConsole object at 0x1175bfee8> to 'Tools' ...
[2019-03-19 15:28:58] - DEBUG -  Adding 'Console' to category 'Tools' ...
DEBUG:typhon.suite:Adding 'Console' to category 'Tools' ...
[2019-03-19 15:28:58] - DEBUG -  Connecting parameter signals ...
DEBUG:typhon.suite:Connecting parameter signals ...
[2019-03-19 15:28:58] - INFO -  Adding devices ...
INFO:typhon.cli:Adding devices ...
```

teddyrendahl:
Logging issues resolved, `QMainWindow` added. Resizing isn't quite right but will be addressed as part of #193

teddyrendahl:
Tests revised, ready for final review.

typhon/tools/log.py
```diff
@@ -10,7 +10,11 @@ class TyphonLogDisplay(TyphonBase):
     """Typhon Logging Display"""
     def __init__(self, level=logging.INFO, parent=None):
         super().__init__(parent=parent)
-        self.logdisplay = PyDMLogDisplay(level=level)
+        # Set the logname to be non-existant so that we do not attach to the
+        # root logger. This causes issue if this widget is closed before the
+        # end of the Python session. For the long term this issue will be
+        # resolved with https://github.com/slaclab/pydm/issues/474
```
klauer: 👍

PR #186, opened 2019-03-14
Child Access through HappiPlugin

### LINKED ISSUE #182: Child Device to fill template in panel
It is reasonable to have a template in which you want to add a child device to an empty Typhon widget. For instance, I want on my main `PIM` screen to have an embedded motor screen for the focus motor component.  You need away to specify these in the template itself. These will not be in `happi` themselves (because the parent is). Maybe you can add this to the channel?

`happi://mfx_dg1_pim.focus` will result in `getattr(happi.load_device('mfx_dg1_pim'), 'focus')`

teddyrendahl:
## Description

If we have a device in `happi` which has a component device that we want a screen for say `reference_laser.y`, we want to be able to grab these devices easily. This PR supports dotted access to components as specified in the channel:

`happi://reference_laser.y`

## Motivation and Context

Closes #182 

## How Has This Been Tested?

Tests added

child_access

### DIFF
--- a/tests/plugins/test_happi.py
+++ b/tests/plugins/test_happi.py
@@ -33,6 +33,16 @@ def test_connection(client):
     hp.remove_connection(hc2)
     assert hp.connections == {}
 
+
+def test_connection_for_child(client):
+    hp = HappiPlugin()
+    mock = Mock()
+    hc = HappiChannel(address='happi://test_motor.setpoint', tx_slot=mock)
+    hp.add_connection(hc)
+    tx = mock.call_args[0][0]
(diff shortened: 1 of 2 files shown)


teddyrendahl:
@klauer Not sure entirely what you mean here but I think `ophyd` will handle multiple layers with `__getattr___` i.e  `getattr(area_detector, 'stats.centroid'). I don't think I guarantee that there aren't `.` in your `ophyd` Device names but I think that is bad enough behavior I don't want to protect against it 😄

klauer:
As mentioned in person, ignore my dumb comment @teddyrendahl...

PR #125, opened 2018-11-21
WIP: rely solely on ophyd metadata subscriptions

klauer:
(WIP, including some other semi-related fixes)

PR #187, opened 2019-03-18
TST: remove fake epics pv

klauer:
Split from #125


klauer:
Passing locally, failing on test_panel_add_enum

klauer:
Removed ophyd-version dependent `FakeEpicsSignal`-enum_strs test. Coverage was confirmed to be identical before/after removal (967 lines missed when only running `test_panel_add_enum`)

PR #188, opened 2019-03-18
ENH: component.doc

### LINKED ISSUE #167: Docstring as Tooltip in TyphonSignalPanel
Use the component docstring as the tooltip in `TyphonSignalPanel`

klauer:
Suggested as a follow-up to #174 by @teddyrendahl 

Use Component docstrings when creating Typhon widgets

Closes #167 

Example screenshot:
image

### DIFF
--- a/typhon/signal.py
+++ b/typhon/signal.py
@@ -16,7 +16,7 @@ from qtpy.QtWidgets import (QGridLayout, QHBoxLayout, QLabel)
 #  Package  #
 #############
 from .utils import (channel_name, clear_layout, clean_attr, grab_kind,
-                    is_signal_ro, TyphonBase)
+                    GrabKindItem, is_signal_ro, TyphonBase)
 from .widgets import (TyphonLineEdit, TyphonComboBox, TyphonLabel,
                       TyphonDesignerMixin, ImageDialogButton,
                       WaveformDialogButton, SignalDialogButton)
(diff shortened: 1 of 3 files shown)


typhon/signal.py
```diff
@@ -293,19 +293,16 @@ def _set_layout(self):
         shown_kind = [kind for kind in Kind if self._kinds[kind.name]]
         # Iterate through kinds
         signals = list()
-        for kind in Kind:
-            if kind in shown_kind:
```
teddyrendahl: Why did I think I needed this? Geez.

typhon/utils.py
```diff
@@ -53,7 +56,7 @@ def is_signal_ro(signal):
 
 
 def grab_kind(device, kind):
-    """Grab all signals of a specific Kind from a Device"""
+    """Grab all signals of a specific Kind from a Device instance"""
```
teddyrendahl: Mind just putting a docstring specifying the returned value? Should have been there before but I figured if I documented them poorly I'd struggle less breaking the API 😄 This seems generally useful outside `typhon`. I like having the `Component` available as well.

typhon/signal.py
```diff
@@ -147,7 +152,7 @@ def add_signal(self, signal, name):
         """
         logger.debug("Adding signal %s", name)
         # Create the read-only signal
-        read = signal_widget(signal, read_only=True)
+        read = signal_widget(signal, read_only=True, tooltip=tooltip)
```
teddyrendahl: So this goes on the read only widget? I kind of imagined it would go on the `QLabel`. That is created in `_add_row`

klauer: Good call. 👍

klauer:
Rebased and force-pushed to remove `fake_epics_pv`; screenshot in PR text.

ZLLentz:
I like this a lot

PR #204, opened 2019-06-25
FIX: certain char array PVs should show as strings

ZLLentz:
## Description

When the signal is noted to be a string (e.g. `Cpt(EpicsSignal, ':TEXT', string=True`), make sure it is displayed as a string in the ui.

## Motivation and Context

Char arrays default to showing the integer values, but these are often used to hold string data and expect the client to translate.

## How Has This Been Tested?

Worked for me, but clearly needs some more rigor.

N/A

### DIFF
--- a/typhon/signal.py
+++ b/typhon/signal.py
@@ -109,7 +109,7 @@ def signal_widget(signal, read_only=False, tooltip=None):
     widget_instance.setObjectName(name)
     if tooltip is not None:
         widget_instance.setToolTip(tooltip)
-    if dtype == 'string':
+    if dtype == 'string' and widget in (TyphonLabel, TyphonLineEdit):
         widget_instance.displayFormat = DisplayFormat.String
     return widget_instance
 
--- a/typhon/signal.py
+++ b/typhon/signal.py
@@ -70,6 +70,7 @@ def signal_widget(signal, read_only=False, tooltip=None):
         desc = {}
     # Unshaped data
     shape = desc.get('shape', [])
+    dtype = desc.get('dtype', '')
     try:
         dimensions = len(shape)
     except TypeError:
@@ -108,11 +109,8 @@ def signal_widget(signal, read_only=False, tooltip=None):
     widget_instance.setObjectName(name)
     if tooltip is not None:
         widget_instance.setToolTip(tooltip)
-    try:
-        if signal.as_string:
-            widget_instance.displayFormat = DisplayFormat.String
-    except AttributeError:
-        pass
+    if dtype == 'string':
+        widget_instance.displayFormat = DisplayFormat.String
     return widget_instance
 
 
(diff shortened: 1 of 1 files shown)


typhon/signal.py
```diff
@@ -107,6 +108,11 @@ def signal_widget(signal, read_only=False, tooltip=None):
     widget_instance.setObjectName(name)
     if tooltip is not None:
         widget_instance.setToolTip(tooltip)
+    try:
```
teddyrendahl: I think this line could be be moved up to line 93. We don't really want to set this on either of the plotting options.

typhon/signal.py
```diff
@@ -107,6 +108,11 @@ def signal_widget(signal, read_only=False, tooltip=None):
...
+        if signal.as_string:
```
teddyrendahl: I'm on the fence here. We could check the `Signal.describe` dictionary to get whether it is a string or not. This would make it less `EpicsSignal` specific. However, this issue seems distinctly specific to `EpicsSignal`. I doubt we will ever have simulated char arrays of strings?

ZLLentz: TBH I forgot that describe was even thing

ZLLentz: `describe` returns a `dict` where `'dtype': 'string'`. Kind of awkward relying on a string containing the text `'string'` but I agree it's more compatible across different signals and worth changing

typhon/signal.py
ZLLentz: Are you saying that you want to prevent these string signals from being plotted?

typhon/signal.py
teddyrendahl: Awkward, but we rely on it in a few other places as well, (mainly in the data plugin), and i believe is required for `bluesky` so I at least expect it to be quite stable.

typhon/signal.py
teddyrendahl: No, the `widget_instance` that is created can either be `PyDMLineEdit` / `PyDMLabel` or a `QPushButton` in the case of arrays. For the latter setting `widget_instance.display_format` does nothing. Where you have it in the current code sets it regardless of the type of `widget_instance` we create.

Sorry this function is a mess. It kind of grew organically and is due for a refactor.

ZLLentz: Got it, though I assume you mean `QComboBox`. I'm assuming this throws an exception then in the case of the combo box?

typhon/signal.py
ZLLentz: Resolving conversation because I changed it

typhon/signal.py
teddyrendahl: I didn't, but that is also true. A `QPushButton` gets created for images and waveforms that allows you to look at the signal in a dialog.

teddyrendahl: This whole block gets simpler if we switch to `PyDMLineEdit` in read-only mode rather than use the label... 🤔

PR #208, opened 2019-07-16
FIX: log the status errors

ZLLentz:
## Description

Sent errors from the status thread to the error logger

## Motivation and Context

Uncaught exceptions in this thread segfault the GUI

## How Has This Been Tested?

It really hasn't been tested enough, just throwing this up for now

### DIFF
--- a/typhon/status.py
+++ b/typhon/status.py
@@ -60,3 +60,5 @@ class TyphonStatusThread(QThread):
         except TimeoutError:
             logger.error("Status %r did not complete in %s seconds",
                          self.status, self.timeout)
+        except RuntimeError as err:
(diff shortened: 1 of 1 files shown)


teddyrendahl:
Don't think this needs testing. Pretty clear we can expect `RuntimeError` from the `wait` function:

https://github.com/bluesky/ophyd/blob/d3cd9b942d1fe1fd7a77355f95e5b7cfe8bb6289/ophyd/status.py#L569

My only concern would be that we need to watch for general exceptions (what if the set method doesn't give us back a valid Status object?), those can be added later however. As long as we are sure that crashing the GUI when a motor fails to move is a bug not a feature this is ready to merge 😄

PR #211, opened 2019-07-19
Use PyPI on Travis

teddyrendahl:
## Description

External collaborators seem to prefer `pip` installing instead of using our CONDA channel. Since both `timechart` and `pydm` are both now on PyPI I decided it was time to add this as a branch in our testing suite and officially support this installation pattern.

Also since Python 3.7 is passing now, I'm removing it from the `allowed_failures` list 🎉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,11 +23,14 @@ matrix:
            - PYDM_CHANNEL=pcds-tag
      - python: 3.6
        env: PYDM_CHANNEL=pydm-dev
+     - python: 3.6
+       env: PYDM_CHANNEL=None
      - python: 3.7
        env:
            - PYDM_CHANNEL=pcds-tag
            - BUILD=1
      - python: 3.7
+       env: PYDM_CHANNEL=None
 
 install:
   - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -1,6 +1,6 @@
-git+https://github.com/slaclab/pydm.git
-git+https://github.com/pcdshub/QDarkStyleSheet.git
```
klauer: Should we nuke our pcdshub/QDarkStyleSheet repo?

.travis.yml
```diff
@@ -23,16 +23,14 @@ matrix:
            - PYDM_CHANNEL=pcds-tag
      - python: 3.6
        env: PYDM_CHANNEL=pydm-dev
+     - python: 3.6
+       env: PYDM_CHANNEL=None
```
klauer: Maybe `PYDM_CHANNEL=pip`? :)

requirements.txt
teddyrendahl: This is probably a good idea.  I will need to change where we build it from in `pcds-recipes` as well.

.travis.yml
```diff
@@ -42,17 +40,31 @@ install:
...
+      source activate test-environment
+    # Install all dependencies via PIP
+    else
+      conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION pip
+      # Launch Conda environment
+      source activate test-environment
+      # Install package
+      python setup.py install
+      # Add our dependencies
+      pip install git+https://github.com/pcdshub/happi.git PyQt5 -r dev-requirements.txt
```
ZLLentz: It looks like you're always testing with happi installed, but not explicitly requiring it for the pip install. Maybe there should be a test env that runs all non-happi tests without happi to make sure it's still ok to omit it.

.travis.yml
```diff
@@ -42,17 +40,31 @@ install:
...
+    # Install all dependencies via CONDA
+    if [[ $PYDM_CHANNEL != "None" ]]; then
+      conda config --add channels $PYDM_CHANNEL
+      # Use the recipe contained inside the repository to build packcage
+      conda build -q conda-recipe --python=$TRAVIS_PYTHON_VERSION --output-folder bld-dir
+      # Install package with development dependencies
+      conda config --add channels "file://`pwd`/bld-dir"
+      conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION typhon --file dev-requirements.txt
+      # Launch Conda environment
+      source activate test-environment
```
ZLLentz: At some point you may consider switching to `conda activate` and setting up the conda environment in the new way.

.travis.yml
teddyrendahl: Fair point. I have this test which checks that everything is importable without `happi` but it might be worth adding a branch here to really test it. With the CONDA install it was awkward because it is listed as a hard requirement in the CONDA recipe so I would have had to uninstall it. However, now that I'm tacking it on to the dev-requirements with pip it would be pretty trivial to not add

https://github.com/pcdshub/typhon/blob/67afff27833471b11cfca6998066ed9b64329ec5/tests/plugins/test_happi.py#L52

.travis.yml
teddyrendahl: Is it as simple as just switch the one word? In that case I'll do it rn

teddyrendahl:
Rebased to clean the history up and added some docs. Ready to merge!

PR #212, opened 2019-08-14
ENH: Python file as a template

### LINKED ISSUE #201: Python Based Template
If we switch to `PyDMEmbeddedWindow` instead of `PyDMDisplay` in `TyphonDeviceDisplay` we can utilize Python based templates.  Requires resolution of https://github.com/slaclab/pydm/issues/481

teddyrendahl:
## Description

This allows a user to specify a Python file instead of a `.ui` as a Typhon template. The Python file must contain a subclass of the `pydm.Display` in accordance with the mandates required by PyDM

@hhslepicka Is this the correct future-proof way of doing this? Seemed pretty easy. Don't know why I was whining so much in the linked issue

I also found a bug where calling `TyphonDeviceDisplay.load_template` without any macros overwrote the cached macros from prior loads which kind of made caching the macros pointless. This is now remedied. 

## Motivation and Context

Closes #201

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,11 +25,11 @@ matrix:
        env: PYDM_CHANNEL=pydm-dev
      - python: 3.6
        env: PYDM_CHANNEL=pip
-     - python: 3.7
+     - python: 3.7.3  # Pinned due to PyDM bug with Python 3.7.4
        env:
            - PYDM_CHANNEL=pcds-tag
            - BUILD=1
-     - python: 3.7
+     - python: 3.7.3  # Pinned due to PyDM bug with Python 3.7.4
        env: PYDM_CHANNEL=pip
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,11 +25,11 @@ matrix:
        env: PYDM_CHANNEL=pydm-dev
      - python: 3.6
        env: PYDM_CHANNEL=pip
-     - python: 3.7.3  # Pinned due to PyDM bug with Python 3.7.4
+     - python: 3.7.3  # Pinned due to PyQt bug. Remove when PyQt 5.13.1 is available
        env:
            - PYDM_CHANNEL=pcds-tag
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -25,11 +25,11 @@ matrix:
        env: PYDM_CHANNEL=pydm-dev
      - python: 3.6
        env: PYDM_CHANNEL=pip
-     - python: 3.7
+     - python: 3.7.3  # Pinned due to PyDM bug with Python 3.7.4
```
hhslepicka: Just a note... it is not a PyDM bug... this was an issue with PyQt that got fixed at 5.13.1 (to be released).
It was holding an exception and Python 3.7.4 is less forgiving of this. Phil Thompson added a bug fix into 5.13.1 but there is no 5.13 at conda-forge, yet...

typhon/display.py
```diff
@@ -139,18 +141,26 @@ def load_template(self, macros=None):
...
-        # Store macros
-        self._last_macros = macros
         try:
-            self._main_widget = Display(ui_filename=self.current_template,
-                                        macros=macros)
+            logger.debug("Loading %s", self.current_template)
+            ext = os.path.splitext(self.current_template)[1]
+            # Support Python files
+            if ext == '.py':
+                logger.debug("Loading as a Python file ...")
```
klauer:
```suggestion
                logger.debug("Loading %r as a Python file ...", self.current_template)
```

.travis.yml
teddyrendahl: Sorry @hhslepicka didn't mean to blame you :smile:

I'll amend the comment.

klauer:
```suggestion
     - python: 3.7.3  # Pinned due to @hhslepicka bug
```

teddyrendahl: :fire:

hhslepicka: source

PR #213, opened 2019-08-20
Automated Template Switching

teddyrendahl:
## Description

The idea was that Typhon would select the correct template based on the size of the widget. This would look really smooth in the docking system. You launch a full display and you get the detailed view. If you tile a bunch they automatically go to the embedded mode, or you can pop it out and expand it to get the engineering screen.

embedded_size

The main issue is some flashing as you get close to a border between templates. The template flips which causes a resize, which causes a template flip which... 💀  I tried adding some hysteresis but didn't get far. 

* This example looks fine, but in the test suite it looks as if multiple templates are being loaded on the same widget? (Investigate with `python run_tests.py tests/test_display --show-ui`)
* I think adding something in the context menu to flip which template you are using could be a worthwhile solution that might be better in the long term.

### DIFF
--- a/tests/test_display.py
+++ b/tests/test_display.py
@@ -43,16 +43,8 @@ def test_device_display(device, motor, qtbot):
     return panel
 
 
-def test_display_without_md(motor, display):
-    # Add a generic motor
-    display.add_device(motor)
-    assert display.devices[0] == motor
-    assert display.current_template == display.templates['detailed_screen']
-
-
 def test_display_with_md(motor, display):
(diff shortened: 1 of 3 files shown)


klauer:
@hhslepicka and I discussed this before, and while we think it's a cool proof-of-concept, it's more likely to confuse users. "No, just resize the window a bit more... the signal should pop up!"

Thanks, SLAC-era @teddyrendahl for the cool idea in any case 😄

PR #219, opened 2019-08-23
PyDM 2.0 and Ophyd Metadata

teddyrendahl:
This is a draft for updating the plugins to be compatible with PyDM 2.0 while also taking advantage of the new `SUB_META` options in Ophyd. This passes will the `master` of both projects. What this doesn't do:

1. This does not work with `Signal`. We depend on the fact that we will receive a connection callback eventually from the signal. For a raw software `Signal` this will never come. We will also never see a value callback if you initialize like `Signal(value=1, ...)`
2. This does not replace the logic where we use the base `ca` plugin for `EpicsSignal`
3. This does not allow a user to subscribe to `SUB_SETPOINT` and `SUB_SETPOINT_META`. The logic would need to watch for `sig://my_signal@SUB_SETPOINT` (exact syntax pending)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -94,13 +94,21 @@ def motor():
 
 class RichSignal(Signal):
 
-    def describe(self):
-        return {self.name: {'enum_strs': ('a', 'b', 'c'),
-                            'precision': 2,
-                            'units': 'urad',
-                            'dtype': 'number',
-                            'shape': []}}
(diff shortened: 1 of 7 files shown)


klauer:
Thanks for the forward-looking attempt here, @teddyrendahl. This branch is now preserved in pcdshub/typhos@sig_md_cb. 

We're still nowhere near getting this implemented, but we can look back here for inspiration whenever that may happen.

PR #222, opened 2019-09-18
DOC: fix exec_ and QApplication args in README

klauer:


teddyrendahl:
Seems to be systematic... person who wrote that must have no idea what they're doing.

PR #229, opened 2020-02-07
WIP: Improvements and fixes

### LINKED ISSUE #189: Recognize Disconnections Faster when adding Device
# Problem
Currently if you add a disconnected `ophyd.Device` to `TyphonSuite`, a display will be created but error messages are barfed out. All calls to the underlying controls system have `try/except` blocks but they don't communicate well. Therefore, we get an error message for every signal that is introspected and every tool we try add a device to.

# Possible Solution
Do we just want to call `wait_for_connection` at the start of the routine and if not just make a "class" based screen for these devices (i.e no information gained via describe?)

### LINKED ISSUE #215: Signal panel sorts seemingly randomly by default
## Expected Behavior

Sorts by order in the class definition

## Current Behavior

Has a different result every time
Screenshot 2019-08-20 at 10 31 16 AM
Screenshot 2019-08-20 at 10 30 40 AM

1. Open the ppm commissioning screen, record the order, repeat

## Your Environment

`typhon=0.4.0a0`
`ophyd=1.3.3`
`pydm=1.7.1`

env is `pcds-2.2.0`

hhslepicka:
## Description

- Add a new widget TyphosLoading which is basically just a loading GIF.
- `add_signal` now only query the `Signal` for metadata if it is connected, otherwise a loading widget is added and replaced with the proper set of widgets when the signal connects. Closes #189
- Fix Signal sorting which when ByKind resulted in screens being displayed in a different order (inside of the Kind) at each time. Closes #215

## Motivation and Context

- TyphosLoading is a good way to provide some feedback for users waiting and it is better than an empty screen
- By not waiting or trying to query the disconnected signal we can provide a better response and experience to users.

## How Has This Been Tested?

Locally

The GIF below shows both changes in action...
typhos_loading


hhslepicka:
Tests are failing because we need a new Happi release.

klauer:
This is awesome - looks even nicer than I had imagined. Will review in a bit.

hhslepicka:
This PR depends on https://github.com/bluesky/ophyd/pull/793 to work with `Signal` and `ophyd.sim` objects.

hhslepicka:
Closing and Reopening PR to kick Travis and get the new Happi v1.3.0 release.

typhos/signal.py
```diff
@@ -154,17 +181,51 @@ def add_signal(self, signal, name, *, tooltip=None):
...
+                    if func.__name__ == '_device_meta_cb':
+                        cid = id
+                        break
+
+                if cid is not None:
+                    signal.unsubscribe(cid)
+
+                # Maybe a HACK to get the _add_devices_cb to happen at the
+                # main thread.
+                method = partial(self._add_devices_cb, name, loc, signal)
```
klauer: Can we use signals/slots instead to keep things in the main thread?

typhos/utils.py
```diff
@@ -102,12 +103,23 @@ def grab_kind(device, kind):
                                 device.configuration_attrs]}[kind]
     # Return that kind filtered for devices
     signals = collections.OrderedDict()
-    device_class = device.__class__
     for attr in kind_attr:
-        signal = getattr(device, attr)
+        if '.' in attr:
+            hierarchy = attr.split('.')
```
klauer: This is a big regression in ophyd if this isn't working - I wonder why... ?

https://github.com/bluesky/ophyd/blob/0c9883548db66b334892140f5875405a5b72a419/ophyd/device.py#L1184

PR #233, opened 2020-02-14
FIX: Too Many Files and Subdevices depth

hhslepicka:
## Description

- I made the stupid mistake of trusting QMovie to close the file, which it clearly doesn't do. Now we cache the loading GIF in the class. This fix the `Too many open files`... sorry @ZLLentz for wasting your time...
- There is no point of showing up all the subdevices in a device tree (e.g. Area Detector with its dozen of subdevices). So we limit it to 1 subdevice level.

### DIFF
--- a/typhos/utils.py
+++ b/typhos/utils.py
@@ -205,12 +205,15 @@ def random_color():
 
 
 class TyphosLoading(QLabel):
+    loading_gif = None
     """Simple widget that displays a loading GIF"""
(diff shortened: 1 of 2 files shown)


typhos/utils.py
```diff
@@ -205,12 +205,15 @@ def random_color():
 
 
 class TyphosLoading(QLabel):
+    loading_gif = None
```
klauer: 🎉 🍷

PR #234, opened 2020-02-15
FIX: Force ophyd.Device to always be lazy_wait_for_connection=False'

hhslepicka:
## Description

Without setting `lazy_wait_for_connection=False` a device that has components which are classified as `lazy` will result in a blocking call when we try to read the `configuration_attrs` and other information as well.
Since on the UI we need to have the asynchronous behavior of the connection, which is different at an interactive shell usage, it is in our best interest to patch the Device class to have `lazy_wait_for_connection=False` by default. That allows us to introspect the inner components and assemble the screen as well as subscribe to callback for proper generation of the UI when this device and its signals connect.

### DIFF
--- a/typhos/__init__.py
+++ b/typhos/__init__.py
@@ -16,3 +16,6 @@ from .plugins import register_signal
 from ._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
+
+from ophyd import Device
+Device.lazy_wait_for_connection = False


klauer:
flake8 failures ☹️ Overall 👍 

As this is an on-import thing, perhaps we want the `Device` laziness patch in our main application (e.g., LUCID) instead of typhos. We might still want to import typhos on the command line and use it, but LUCID would be less likely. Thoughts?

hhslepicka:
You are right... maybe I should apply time into making this a context manager thing that we can use with Typhos or LUCID code and revert when we are done? Thoughts? 
In any case I think we are better moving this PR to LUCID as you mentioned.

The context manager would be the same idea as what you have at Ophyd with the do_not_wait... but applied to Device class in order to avoid the issue with subdevices.

Thank you for reviewing it

hhslepicka:
Closed in favor of https://github.com/pcdshub/lucid/pull/25

PR #239, opened 2020-02-24
ENH: Use exponential displayFormat if EGU requires it.

hhslepicka:
## Description

This PR tries to be a bit smart on using an adequate displayFormat depending on the kind of information we are displaying. This was caught by @slacAdpai while testing an expert screen generated via `Typhos` in which the displayed value did not match the display at the `pcdswidgets` readback. That was due to the `displayFormat` that was set to be `Exponential` in one and `Default` at the other. With this PR we try to be a bit smart at the autogeneration and rely at the EGU information coming from the control system to switch.

## Motivation and Context

Engineers and scientists would like to properly see the value of the readings.

## How Has This Been Tested?

Locally

### DIFF
--- a/typhos/widgets.py
+++ b/typhos/widgets.py
@@ -23,6 +23,8 @@ from pyqtgraph.parametertree import parameterTypes as ptypes
 
 logger = logging.getLogger(__name__)
 
+EXPONENTIAL_UNITS = ['mtorr', 'torr', 'kpa', 'pa']
+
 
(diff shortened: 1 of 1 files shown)


hhslepicka:
Closes https://github.com/pcdshub/pcdswidgets/issues/30

hhslepicka:
Depends on https://github.com/slaclab/pydm/pull/597

PR #240, opened 2020-02-24
FIX: Make sure that the stylesheet is not affecting the widgets that are not Typhos

hhslepicka:
## Description

The Typhos stylesheet was not restricting the selection to descendants of TyphosBase but instead it was affecting other widgets with its rules.
This PR restricts the selection to descendants of TyphosBase.

## Motivation and Context

Other screens being launched from an application that applied the Typhos stylesheet were looking wrong and having a completely wrong appearence.

## How Has This Been Tested?

Locally

### DIFF
--- a/typhos/ui/style.qss
+++ b/typhos/ui/style.qss
@@ -1,84 +1,84 @@
-QWidget{
+TyphosBase {
    background-color: #ffffff;
    color: black;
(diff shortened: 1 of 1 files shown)


hhslepicka:
The current screens behave as expected.

PR #241, opened 2020-02-25
REF: Move no_device_lazy_load from lucid to typhos.

hhslepicka:
TST: Add tests for typhos.utils.no_device_lazy_load.

## Description

It makes more sense for this function to live here.


hhslepicka:
Now this should work and not take forever to open due to every signal timeout on connection:
```python
from qtpy.QtWidgets import QApplication
import typhos
app = QApplication([])
import pcdsdevices.spectrometer
with typhos.utils.no_device_lazy_load():
    mono = pcdsdevices.spectrometer.Kmono('PREFIX:', name='mono')
    display = typhos.TyphosSuite.from_device(mono)
display.show()
```

PR #245, opened 2020-02-26
ENH: New typhos cli functionality

hhslepicka:
## Description

ENH: Add option at typhos.cli to allow for users to pass a class and args.
ENH: Added TyphosDeviceDisplay.from_class to allow a display to be created via a class and args/kwargs.
BLD: Added dependency for pcdsutils.
TST: Added tests for the new cli functionality.

## Motivation and Context

@klauer mentioned that it would be really useful for engineers and users in general to be able to quickly check a screen for a particular device.
Now the `typhos` command line interface allow that by passing a class and kwargs in the following format:
```bash
$ typhos "package.ClassName[{"arg1":"val1",...}]"
```
The command line interface still support Happi entries and they can be mixed together.

## How Has This Been Tested?

New tests added for `typhos.cli`.

argparse help was expanded to also describe the possibility to load from a class.

## Dependencies
This PR depends on https://github.com/pcdshub/pcdsutils/pull/4

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -19,6 +19,7 @@ requirements:
       - happi >=1.3.0
       - numpy
       - ophyd >=1.2.0
+      - pcdsutils
       - pyqtgraph
       - pydm >=1.6.0
       - qdarkstyle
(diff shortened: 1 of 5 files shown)


hhslepicka:
Tests are passing locally. I am not sure why it says that it can't find the `no_device_lazy_load` function inside of the package itself.

Issue #248, opened 2020-03-05
Typhos Template Discussion/Implementation


ZLLentz:
See https://github.com/pcdshub/pcdsdevices/pull/382

klauer:
Copying my comment here for further discussion - please copy in a summary of your thoughts as well.

Order of template determination
-------------
1. User/developer -level override somehow?
2. `happi` entry template field
3. Class/instance-specified templates
   a. Fully qualified class path, e.g., `{pcdsdevices.whatever.ClassName}.{detailed}.ui`
   b. Bare class path, e.g., `{ClassName}.{detailed}.ui`
   c. Repeat the above for `cls.mro()[1:]`

Other
-------------
~* Template file searching should likely be cached~
* I also wonder if we could have any of the above alternate templates for screens swapped dynamically with, say, a context menu option. This could allow for the operator to fall back to a screen easily if we incorrectly specify it somehow.

ZLLentz:
Adding:

- Use class `mro` to pick a fallback screen, e.g. positioner subclasses can default to the `ophyd.positioner.PositionerBase.ui` file
- Provide option in `TyphosSignalPanel` to omit signals that are specified elsewhere in the template

hhslepicka:
The context menu option is a great idea. I will try to take a look at it.

klauer:
I like the idea of:
>Provide option in TyphosSignalPanel to omit signals that are specified elsewhere in the template, e.g. no reason to show the motor's readback in the middle of the big sub-tab if we have it front and center above

I think it's much more complicated than I had first imagined. I'd like to move it to a separate issue and consider this one closed once #257 is in.

ZLLentz:
I imagine it would require a bit of a refactor to accomplish that item... I think it's a good thing to work towards but maybe not the highest priority

klauer:
Thanks for the input, @ZLLentz. Closing.

PR #249, opened 2020-03-05
FIX: restore missing show signals tags

ZLLentz:
## Description

In most of the default templates, some of the `TyphosSignalPanel` instances were missing various kind show tags, mostly for omitted signals but in some cases other types were missing as well. These properties had checkboxes in designer but I have observed that the `typhos` screens are being loaded without the omitted signals. I unchecked and rechecked the boxes and pressed save.

## Motivation and Context

Maybe this will make the omitted signals come back? Probably not.

### DIFF
--- a/typhos/ui/detailed_positioner.ui
+++ b/typhos/ui/detailed_positioner.ui
@@ -349,6 +349,9 @@
              <property name="showConfig" stdset="0">
               <bool>false</bool>
              </property>
+             <property name="showOmitted" stdset="0">
+              <bool>true</bool>
+             </property>
             </widget>
            </item>
           </layout>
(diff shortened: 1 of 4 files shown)


hhslepicka:
@ZLLentz I don't think we should change the meaning of `omitted`. Otherwise what would we use for signals that we really want to omit?

ZLLentz:
@hhslepicka I guess I don't understand what goes in the `miscellaneous` tab then. I feel like we've had this conversation before and it doesn't stick.

ZLLentz:
@hhslepicka easily convinced me in person that omitted signals should be left out everywhere, because they are omitted. I opened #250 based on our in-person discussion plus some ideas that were floating in my head. I also added one of the ideas from this discussion to #248.

Closing because this should be handled a different way and the issues are opened to plan a path.

PR #255, opened 2020-03-07
CLN/STY: isort it all (import sorting)

klauer:


ZLLentz:
I'm really happy that `isort` is a thing, TIL

PR #256, opened 2020-03-07
ENH: Add context menu to switch template.

hhslepicka:
## Description

Provide users a way to change the current `template` in use via a Context Menu.

## Motivation and Context

As we move ahead with the development of Templates and even the ability to specify templates based on the class hierarchy of devices it is desirable to have the option to pick which template best fits the need for a certain task.

## How Has This Been Tested?

Locally

N/A

template_switch

### DIFF
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -2,9 +2,10 @@ import enum
 import logging
 import os.path
 import pathlib
+import functools
 
-from qtpy.QtCore import Q_ENUMS, Property, Slot
-from qtpy.QtWidgets import QHBoxLayout, QWidget
+from qtpy.QtCore import Q_ENUMS, Property, Slot, Qt
+from qtpy.QtWidgets import QHBoxLayout, QWidget, QMenu
 
 import pcdsutils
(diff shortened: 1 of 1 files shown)


ZLLentz:
That is really neat

typhos/display.py
```diff
@@ -80,6 +81,43 @@ def __init__(self, parent=None, **kwargs):
...
+        QMenu
+        """
+        def switch_template(fname):
+            self.force_template = fname
+
+        menu = QMenu(parent=self)
+
+        for display_name, filename in self.templates.items():
+            action = menu.addAction(display_name)
+            action.triggered.connect(functools.partial(switch_template,
```
klauer: I believe if you put the `switch_template` in the `for` loop, you can use `filename` as-is:

```suggestion
            def switch_template():
                self.force_template = filename
            action.triggered.connect(switch_template)
```

hhslepicka:
@klauer @ZLLentz this is good to go...
Feel free to provide suggestions, feedback, requirements...

PR #257, opened 2020-03-11
ENH: entirely rework display loading mechanism, add display switcher

klauer:
Having quick access to switching displays I think will be a useful thing:

image

Left-click for a quick swap to the "best" embedded/engineering/detailed display.
Right-click to choose a specific embedded/engineering/detailed display.

image

Alternatively, the display context menu was tweaked a bit:

image

I think we can do better widget/style/etc. wise, so consider this a proof-of-concept.

This PR also entirely reworks how displays are searched for and loaded.

TODO
- [x] Update paths to search to include the pydm display path
- [x] Fix tests
- [x] Make buttons nicer or change them entirely based on feedback (@hhslepicka @ZLLentz?)
- [x] Include the switcher in the remaining default templates
- [x] Add icons
- [x] Fix pydm deprecation warning for display loading
- [ ] Rename positioner templates such that positioner class MRO will pick these templates by default

### DIFF
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -9,6 +9,7 @@ from qtpy.QtCore import Q_ENUMS, Property, Slot, Qt
 import pcdsutils
 import pcdsutils.qt
 import pydm.display
+import pydm.exception
 import pydm.utilities
 
 from . import utils
@@ -419,10 +420,16 @@ class TyphosDeviceDisplay(utils.TyphosBase, widgets.TyphosDesignerMixin,
         else:
             try:
                 self._load_template(template)
-            except Exception:
+            except Exception as ex:
                 logger.exception("Unable to load file %r", template)
-                self._main_widget = QtWidgets.QWidget()
-                self._current_template = None
+                # If we have a previously defined template
+                if self._current_template is not None:
+                    # Fallback to it so users have a choice
+                    self._load_template(self._current_template)
+                    pydm.exception.raise_to_operator(ex)
+                else:
+                    self._main_widget = QtWidgets.QWidget()
+                    self._current_template = None
 
         self.layout().addWidget(self._main_widget)
         utils.reload_widget_stylesheet(self)
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -457,16 +457,16 @@ class TyphosDeviceDisplay(utils.TyphosBase, widgets.TyphosDesignerMixin,
 
         self._main_widget = main = loader(str(filename), macros=self._macros)
 
-        # Add device to all children widgets
-        if not self.devices:
-            return
+        self._main_widget = main = loader(str(filename), macros=self._macros)
 
+        # Notify child widgets of: this device dispay + the device
+        device = self.device
         designer = main.findChildren(widgets.TyphosDesignerMixin) or []
         bases = main.findChildren(utils.TyphosBase) or []
 
-        device, = self.devices
         for widget in set(bases + designer):
-            widget.add_device(device)
+            if device and hasattr(widget, 'add_device'):
+                widget.add_device(device)
 
             if hasattr(widget, 'set_device_display'):
                 widget.set_device_display(self)
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -143,8 +143,6 @@ class TyphosDisplayTitle(QtWidgets.QFrame, widgets.TyphosDesignerMixin):
         self._show_switcher = show_switcher
         super().__init__(parent=parent)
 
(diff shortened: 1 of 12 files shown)


klauer:
Added some icons with pydm/@hhslepicka's help:

image

klauer:
I may have messed up the templates a bit - someone who knows them better should take a look.

hhslepicka:
Wicked! I really liked the icons and how it looks like. I will take it for a spin and let you know.

typhos/display.py
```diff
@@ -61,25 +163,33 @@ class TyphosDeviceDisplay(TyphosBase, TyphosDesignerMixin, _DisplayTypes):
...
-            for templ in self.TemplateEnum
+        self.templates = {type_.name: [] for type_ in DisplayTypes}
+
+        instance_templates = {
+            'embedded_screen': embedded_templates or [],
+            'detailed_screen': detailed_templates or [],
+            'engineering_screen': engineering_templates or [],
         }
+        for view, path_list in instance_templates.items():
+            paths = [pathlib.Path(p).expanduser().resolve() for p in path_list]
```
hhslepicka: Maybe also expand env vars?

typhos/display.py
```diff
@@ -182,54 +302,73 @@ def load_template(self, macros=None):
...
+        macros = macros or self._macros
+        ret = {}
+        for display_type in DisplayTypes:
+            ret[display_type.name] = None
+            try:
+                value = self._macros[display_type]
+            except KeyError:
+                ...
+            else:
+                value = pathlib.Path(value).expanduser().resolve()
```
hhslepicka: Same as above about the expand vars

typhos/utils.py
```diff
@@ -22,11 +22,25 @@
...
-ui_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'ui')
+MODULE_PATH = pathlib.Path(__file__).parent.resolve()
+ui_dir = MODULE_PATH / 'ui'
 GrabKindItem = collections.namedtuple('GrabKindItem',
                                       ('attr', 'component', 'signal'))
 
 
+def _get_display_paths():
+    'Get all display paths based on PYDM_DISPLAY_PATHS + typhos built-ins'
+    paths = os.environ.get('PYDM_DISPLAYS_PATH', '')
```
hhslepicka: Good idea to reuse the PYDM_DISPLAYS_PATH.

typhos/display.py
klauer: You have unexpanded environment variables in `PYDM_DISPLAY_PATH`? I'd expect bash would handle that prior to running typhos.

Or are you thinking of macros in the path list?

typhos/display.py
```diff
@@ -26,9 +25,202 @@ class DisplayTypes(enum.IntEnum):
...
+class TyphosDisplayTitle(QtWidgets.QFrame, widgets.TyphosDesignerMixin):
+    """
+    Standardized Typhos Device Display title
+    """
+    def __init__(self, title='${name}', *, show_switcher=True, parent=None):
+        self._show_switcher = show_switcher
+        super().__init__(parent=parent)
+
+        font = QtGui.QFontDatabase.systemFont(QtGui.QFontDatabase.TitleFont)
+        font.setPointSizeF(14.0)
```
hhslepicka: Can we change the size to be in pixel size? According to https://doc.qt.io/qt-5/highdpi.html it will be an issue due to specification in PointSize.

hhslepicka:
All the features that this PR adds are great. Thanks a lot for all the effort, @klauer. 🅰️ ➕ 
I will start adding here the things I find and we can decide how we split the work to address them.

- When using via lucid LFE, if I open a device such as `im4l0` all looks good initially, after pressing the `embedded` view, it compress everything. If I switch to another template all looks good. Going back to `embedded` results in a blank screen followed by a long delay and the screen shows again.
Here is a GIF demonstrating this behavior:
typhos_template_switch

klauer:
Do you think before switching displays, we should stash a dictionary of `{current_template: current_display_size}` and restore that by default?

hhslepicka:
No need if we set the proper flags for the sizePolicy... I believe that MinimumExpanding for both should address this issue. Otherwise the screens will not be able to resize.

typhos/display.py
klauer: Feel free to commit this change

hhslepicka: I will get it in

typhos/display.py
klauer: @hhslepicka noted in private this isn't an issue and worked locally

PR #265, opened 2020-04-13
REF/FIX: plot rework + background connections

### LINKED ISSUE #194: Plot Attempts to add hinted devices
This leaves the user with a logger exception like:

```python
[2019-03-19 22:22:20] - ERROR -  Unable to add ycenter to plot-able signals
  File "/reg/g/pcds/pyps/conda/py36/envs/pcds-2.1.0/lib/python3.6/site-packages/typhon/tools/plot.py", line 146, in add_device
    if sig.describe()[sig.name]['dtype'] in ('integer',
KeyError: 'mfx_dg1_slits_ycenter'
```

klauer:
Closes #194

* Adds a suite of utilities for dealing with ophyd subscriptions in a temporary sort of way (with context managers)
* Adds a QThread wrapper that emits events for connections
* Hacks on some signal-handling to see if a `Signal` might not emit a connection callback (because ophyd is busted)
* Limits number of channels automatically added as hints
* Uses fully-qualified names (root.name + `dotted_name`) for signals added to the plot

What else, @hhslepicka, do we need here?

### DIFF
--- a/typhos/tools/plot.py
+++ b/typhos/tools/plot.py
@@ -3,18 +3,18 @@ Typhos Plotting Interface
 """
 import logging
 
-from ophyd import Device
+from qtpy.QtCore import Slot
+from qtpy.QtWidgets import (QComboBox, QHBoxLayout, QLabel, QPushButton,
+                            QVBoxLayout)
 from timechart.displays.main_display import TimeChartDisplay
 from timechart.utilities.utils import random_color
-from qtpy.QtCore import Slot
-from qtpy.QtWidgets import (QComboBox, QPushButton, QLabel, QVBoxLayout,
-                            QHBoxLayout)
-from ..utils import (channel_from_signal, clean_attr, clean_name, TyphosBase)
+
+from .. import utils
 
 logger = logging.getLogger(__name__)
 
 
-class TyphosTimePlot(TyphosBase):
+class TyphosTimePlot(utils.TyphosBase):
     """
     Generalized widget for plotting Ophyd signals
 
@@ -44,6 +44,7 @@ class TyphosTimePlot(TyphosBase):
         self.timechart = TimeChartDisplay(show_pv_add_panel=False)
         self.layout().addWidget(self.timechart)
         self.channel_map = self.timechart.channel_map
+        self._device_threads = {}
 
     def add_available_signal(self, signal, name):
         """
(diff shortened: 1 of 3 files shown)


typhos/utils.py
```diff
@@ -517,3 +519,200 @@ def find_file_in_paths(filename, *, paths=None):
...
+        Filter signals, with signature ``callable(ophyd.Device.ComponentWalk)``
+    '''
+    signals = get_all_signals_from_device(device, include_lazy=include_lazy)
+    with subscription_context(*signals, callback=callback,
+                              event_type=event_type, run=run) as obj_to_cid:
+        yield obj_to_cid
+
+
+class _ConnectionStatus:
+    def __init__(self, signals, callback):
```
ZLLentz, line 616: `signals` is an unused argument here

klauer: Good catch - fixed in 35c850d. 

Do you think the class itself adds any value? That is, it's yielded by the context manager and summarizes connection status. It could be up to the caller to keep track of it.

ZLLentz: I can't imagine any specific use case, but it seems useful in abstract

klauer:
There was an attempt to fix the ophyd ugliness here: https://github.com/bluesky/ophyd/pull/793
But it wasn't exactly the "right" solution (which I have a feeling will be much more complicated).

typhos/utils.py
klauer: OK 👍 

Hmm, maybe it could gain a `remove_signal` method in the future, or something along those lines. We'll see how it's used and go from there.

typhos/tools/plot.py
```diff
@@ -120,33 +130,42 @@ def creation_requested(self):
...
-                            if sig.name in subdevice.hints.get('fields', []):
-                                self.add_curve(channel_from_signal(sig),
-                                               name=name)
-                    except Exception:
-                        logger.exception("Unable to add %s to "
-                                         "plot-able signals",
-                                         component)
+
+        if device not in self._device_threads:
+            thread = utils.DeviceConnectionMonitorThread(
```
hhslepicka, line 167: 👏 This will be used an awful lot.

klauer:
Thanks for the reviews guys!

PR #268, opened 2020-04-14
ENH: add setpoint history context menu entry

### LINKED ISSUE #226: Drop-down history for (positioner) setpoint requests
Show history as in:
> (Timestamp) (Previous value1)
> (Timestamp) (Previous value2)
> (Timestamp) (Previous value3)

Action on clicking - set textbox, do not move (?)

klauer:
Closes #226

Originally this started as a drop-down for setpoint line edits, but I think it might fit more neatly in the context menu as pydm already modifies it:

image

Functionality:
* Records entries as the user adds them to the item, along with a timestamp.
* Displays history in context menu
* Limits history to a designable Property value (0 to disable entirely)
* Tweaks the menu font to allow for visual alignment of timestamp/value

### DIFF
--- a/tests/test_widgets.py
+++ b/tests/test_widgets.py
@@ -1,8 +1,9 @@
 import pytest
-import qtawesome as qta
-from qtpy.QtGui import QIcon
 from qtpy.QtWidgets import QWidget
 
+import ophyd
+import pydm
+from typhos import widgets
 from typhos.suite import SidebarParameter
 from typhos.widgets import (ImageDialogButton, QDialog, SignalDialogButton,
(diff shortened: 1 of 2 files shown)


klauer:
Thanks! I think/hope it could go in PyDM eventually.

hhslepicka:
@ZLLentz I think that this feature should be on the timeline for 2.0 for sure. I would also like to investigate for 2.0 something that @klauer worked on a while ago which was a mouse tooltip with the last X values of the widget in form of a Timeplot.
For now I think we can mature this feature at Typhos and bring that into PyDM later on.

PR #277, opened 2020-04-25
ENH: Allow SignalDialogButton to produce persistent widgets.

hhslepicka:
## Description
Add a class attribute `persistent` to `SignalDialogButton` allowing us to configure wether or not the dialog generated to display the widget will be available even if the parent is destroyed or not.

## Motivation and Context
This change allow us to have images and plots as standalone dialogs even if we change the layout or if we close the current screen for a device.
That is useful when performing diverse operations.

## How Has This Been Tested?
Locally.

N/A

persistent

### DIFF
--- a/typhos/utils.py
+++ b/typhos/utils.py
@@ -863,3 +863,23 @@ def _get_top_level_components(device_cls):
     Get all top-level components from a device class
     """
     return list(device_cls._sig_attrs.items())
+
+
+def find_parent_with_class(widget, cls=QWidget):
+    """
+    Finds the first parent of a widget that is an instance of ``klass``
+
+    Parameters
+    ----------
+    widget : QWidget
+        The widget from which to start the search
+    cls : type, optional
+        The class which the parent must be an instance of
+
+    """
+    parent = widget
+    while parent is not None:
+        if isinstance(parent, cls):
+            return parent
+        parent = parent.parent()
+    return None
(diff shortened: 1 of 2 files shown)


hhslepicka:
The only downside of this is that if we close the main window, the dialogs still remain open. I could not find a good way to hook it up to close all. I thought that hooking it up with the Suite could limit the usage of the widgets. 
Any thoughts/suggestions?

ZLLentz:
Is there a way to set `parent` to be the main window instead of setting `persistent` to `True`? I suspect that this would cause the dialogs to remain open as long as the application is open.

hhslepicka:
@ZLLentz 
Setting it to the main window will help with closing it and eliminating the window, the problem becomes which main window to select? The first instance of QMainWindow that we can find looping over widget.parent() ?

Another option would be to look for the first instance of `TyphosSignalPanel` and try to set that as the parent but it would mean that if someone closes that device panel the floating window will be gone.

klauer:
Yeah, I think it makes sense to look back up the widget hierarchy to find its parent. 

For maximum generality, it could be a class-defined `parent_widget_class` that says "look through my ancestry to find this class type", perhaps?

hhslepicka:
@klauer @ZLLentz I refactored the code to use the suggested parent class for search.
This is good for another review.

PR #279, opened 2020-04-27
Path caching in `find_templates_for_class`

klauer:
master baseline on NFS: 

```
In [1]: %timeit list(typhos.utils.find_templates_for_class(ophyd.EpicsMotor, 'embedded', paths=typhos.utils.DISPLAY_PATHS))
2.93 ms ± 12.5 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

This PR:

```
In [2]: %timeit list(typhos.utils.find_templates_for_class(ophyd.EpicsMotor, 'embedded', paths=typhos.utils.DISPLAY_PATHS))
1.08 ms ± 15.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

This is called 3x per display, and each call time scales linearly with `len(cls.mro())`

So assuming 200 devices, baseline should be 200 * 3 * 3ms = 1.8sec
With this, 200 * 3 * 1ms = 0.6s
Moving the caching up a level as in 18dbeb5, should reduce this by ~3

Not sure it's worth the complexity just yet

Alternatives
-------------
* LRU cache (builtin) - and a custom time-to-live / stale threshold https://docs.python.org/3/library/functools.html#functools.lru_cache
* 3rd party library https://cachetools.readthedocs.io/en/stable/

### DIFF
--- a/typhos/cache.py
+++ b/typhos/cache.py
@@ -228,7 +228,9 @@ class _GlobalWidgetTypeCache(QtCore.QObject):
 
 
 # The default stale cached_path threshold time, in seconds:
-TYPHOS_PATH_CACHE_TIME = int(os.environ.get('TYPHOS_PATH_CACHE_TIME', '600'))
+TYPHOS_DISPLAY_PATH_CACHE_TIME = int(
+    os.environ.get('TYPHOS_DISPLAY_PATH_CACHE_TIME', '600')
+)
 
 
 class _CachedPath:
@@ -253,7 +255,8 @@ class _CachedPath:
         happens on the next glob, and not on a timer-basis.
     """
 
-    def __init__(self, path, *, stale_threshold=TYPHOS_PATH_CACHE_TIME):
+    def __init__(self, path, *,
+                 stale_threshold=TYPHOS_DISPLAY_PATH_CACHE_TIME):
         self.path = pathlib.Path(path)
         self.cache = None
         self._update_time = None
@@ -334,6 +337,7 @@ class _GlobalDisplayPathCache:
             The path to add.
         """
         path = pathlib.Path(path).expanduser().resolve()
-        path = _CachedPath(path, stale_threshold=TYPHOS_PATH_CACHE_TIME)
+        path = _CachedPath(
+            path, stale_threshold=TYPHOS_DISPLAY_PATH_CACHE_TIME)
         if path not in self.paths:
             self.paths.append(path)
--- a/typhos/cache.py
+++ b/typhos/cache.py
@@ -1,8 +1,14 @@
+import fnmatch
 import functools
 import logging
-import ophyd
+import os
+import pathlib
+import re
+import time
 
 from qtpy import QtCore
+
+import ophyd
 
 from . import utils
 from .widgets import SignalWidgetInfo
@@ -14,6 +20,7 @@ logger = logging.getLogger(__name__)
 # `get_global_describe_cache()` and `get_global_widget_type_cache()` below.
 _GLOBAL_WIDGET_TYPE_CACHE = None
 _GLOBAL_DESCRIBE_CACHE = None
+_GLOBAL_PATH_CACHE = None
 
 
 def get_global_describe_cache():
@@ -30,6 +37,14 @@ def get_global_widget_type_cache():
     if _GLOBAL_WIDGET_TYPE_CACHE is None:
         _GLOBAL_WIDGET_TYPE_CACHE = _GlobalWidgetTypeCache()
     return _GLOBAL_WIDGET_TYPE_CACHE
+
+
+def get_global_path_cache():
+    """Get the _GlobalDisplayPathCache singleton."""
+    global _GLOBAL_PATH_CACHE
+    if _GLOBAL_PATH_CACHE is None:
+        _GLOBAL_PATH_CACHE = _GlobalDisplayPathCache()
+    return _GLOBAL_PATH_CACHE
 
 
 class _GlobalDescribeCache(QtCore.QObject):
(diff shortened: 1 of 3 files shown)


klauer:
The impact of application-level path caching is greater than I had previously estimated. I think it's time we move forward with this.

For the sake of numbers and benchmarking, we can disable the caching by changing the environment variable `TYPHOS_PATH_CACHE_TIME`.

typhos/cache.py
```diff
@@ -210,3 +225,112 @@ def get(self, obj):
...
+    def __hash__(self):
+        # Keep the hash the same as the internal path for set()/dict() usage
+        return hash(self.path)
+
+    @property
+    def time_since_last_update(self):
+        """Time (in seconds) since the last update."""
+        if self._update_time is None:
+            return 0
+        return time.monotonic() - self._update_time
```
ZLLentz, line 286: This is a new one for me. Today I learned I should be using `time.monotonic()` to be robust against OS clock screwiness.

typhos/cache.py
```diff
@@ -210,3 +225,112 @@ def get(self, obj):
...
+                    yield self.path / path
+        else:
+            # No globbing syntax: only check if file is in the list
+            if pattern in self.cache:
+                yield self.path / pattern
+
+
+class _GlobalDisplayPathCache:
+    """
+    A cache for all configured display paths
```
ZLLentz, line 315: Our new style guide says that docstrings should end in periods

typhos/cache.py
```diff
@@ -32,6 +39,14 @@ def get_global_widget_type_cache():
     return _GLOBAL_WIDGET_TYPE_CACHE
 
 
+def get_global_path_cache():
+    """Get the _GlobalDisplayPathCache singleton."""
+    global _GLOBAL_PATH_CACHE
+    if _GLOBAL_PATH_CACHE is None:
+        _GLOBAL_PATH_CACHE = _GlobalDisplayPathCache()
```
ZLLentz, line 46: There are some inconsistencies here and there between calling it a "global display path cache" vs a "global path cache", this tripped me up a bit when trying to read through the code paths

typhos/display.py
```diff
@@ -781,6 +782,7 @@ def display_widget(self):
     @staticmethod
     def _get_templates_from_macros(macros):
         ret = {}
+        paths = [cache.get_global_path_cache()]
```
ZLLentz, line 785: Something seems strange here. This is a list of one item, the path cache, and it is being passed into `utils.find_file_in_paths` in place of what is expected to be an iterable of paths.

typhos/cache.py
```diff
@@ -210,3 +225,112 @@ def get(self, obj):
             if desc is not None:
                 # In certain scenarios (such as testing) this might happen
                 self._new_description(obj, desc)
+
+
+# The default stale cached_path threshold time, in seconds:
+TYPHOS_PATH_CACHE_TIME = int(os.environ.get('TYPHOS_PATH_CACHE_TIME', '600'))
```
hhslepicka, line 231: I like the idea of having the default value and the opt-out aspect of it since this addition brings much value.

typhos/cache.py
klauer: Resolved in 6e51a31be01f94cb31d6b0cbd58cb4e5157aece1

typhos/cache.py
klauer: Resolved in 6e51a31be01f94cb31d6b0cbd58cb4e5157aece1

typhos/cache.py
```diff
@@ -210,3 +225,120 @@ def get(self, obj):
...
```
hhslepicka, line 231: Maybe we should switch this envvar name to be explicit about the PATH being DISPLAYS_PATH?

typhos/display.py
klauer: I wanted an escape-hatch from the path deduplication in `find_file_in_paths`, which is how this implementation resulted. 

Alternative in: c993a1b

typhos/cache.py
klauer: I'm happy to change it if you can suggest a name without making it extremely long to type?

hhslepicka: I was going to suggest the addition of displays in the name but it will be long... feel free to skip this one here.
`TYPHOS_DISPLAYS_PATH_CACHE_TIME` is ginormous

typhos/utils.py
```diff
@@ -493,13 +493,10 @@ def find_templates_for_class(cls, view_type, paths, *, extensions=None,
...
-    if len(paths) > 1:
-        paths = remove_duplicate_items(
-            [pathlib.Path(p) for p in paths]
-        )
-
     from .cache import _CachedPath
-    paths = [_CachedPath.from_path(p) for p in paths]
+    paths = remove_duplicate_items(
+        [_CachedPath.from_path(p) for p in paths]
+    )
```
hhslepicka: Could we move the `remove_duplicate_items` to the `_GlobalDisplayPathCache.add_path`?
So far this function is invoked only at `typhos/display.py:1040` and there `paths` is passed as `paths = cache.get_global_display_path_cache().paths`
That would allow us to do the search for templates without realocation and sorting and cast and only perform the operation when new paths are added. Would that work?

klauer:
I think a persistent cache is compelling when:
1. Information takes a long time to acquire (connect to PV, `describe` + `get`)
2. Information may be periodically unavailable (IOC is down)

So putting that together:
1. NFS is slow, but storing a list of files and reading it from a database on NFS is probably slower
2. If the directories are unavailable, so are the `.ui` files

typhos/utils.py
```diff
@@ -536,13 +533,10 @@ def find_file_in_paths(filename, *, paths=None):
...
```
hhslepicka: I believe this to be the same case as above in which we already pass the paths as `paths = cache.get_global_display_path_cache().paths` but I am not sure if that will always be true.

typhos/utils.py
klauer: It'd work, but it would remove the overall generality of the utility function which I've been trying to retain for future usage 🤷

typhos/cache.py
klauer: Tie breaker @ZLLentz?

typhos/utils.py
hhslepicka: Sounds good! 👍

hhslepicka:
Looking at the alternatives, I like your cache implementation better than LRU cache as the LRU seems to be dependent of a function call and we would keep two caches for the two find functions. 
For the `cachetools`, your implementation looks similar to the cachetools.TTLCache, I don't think that we need another dependency for that.

typhos/cache.py
ZLLentz: I think it's better to be clear than succinct (of course being both of these is best, but...)

`TYPHOS_DISPLAY_PATH_CACHE_TIME` is my vote, but I don't have a strong opinion here since I don't expect this to be changed often.

klauer: Resolved in 487f98e

PR #280, opened 2020-04-28
WIP: Hide empty panels option.

### LINKED ISSUE #278: Hide Empty Panels at Composite Signal Panel
Transitions between embedded/detailed/engineering is slick, even for the monster motors/states/area detector device that is the PPM. The area detector section in particular is a very good best-effort generation and really makes it clear where the device's kinds needed to be fixed. I also really like the context menu where I can filter on signal kind.

One issue: empty  filtered composite panels still linger. Shown here, PPM with only hinted/read available:

config_linger

_Originally posted by @ZLLentz in https://github.com/pcdshub/typhos/pull/269#issuecomment-618575045_

hhslepicka:
## Description
Provide option to hide empty panels.
Closes #278 

## Motivation and Context
Discussion at PR #278 

## How Has This Been Tested?
Locally

This almost works for all cases... I am now having an issue with a situation in which `TyphosDeviceDisplay` is always visible if it was more `TyphosDeviceDisplay` in it.
I need some help to understand how to deal with this situation.

### DIFF
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -1,4 +1,5 @@
 import enum
+import functools
 import logging
 import os.path
 import pathlib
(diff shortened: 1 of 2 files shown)


hhslepicka:
Closing in favor of #284

Issue #271, opened 2020-04-23
CLI Full Class Edge Case


ZLLentz:
## Expected Behavior

Should be able to load devices like `typhos "pcdsdevices.device_types.PPM[{'prefix':'IM3L0:PPM', 'name':'im3l0'}]"` without a `happi` configuration

## Current Behavior

If `happi` is installed and no `happi` config is found, there will be an error raised regardless of whether I asked for a `happi` name or a class designation as above

## Possible Solution

Locally I removed this line to help test the current PR: https://github.com/pcdshub/typhos/blob/1e4bb1bcd6c7b0165e3cedda47d9c4e42c900b9e/typhos/cli.py#L92
This isn't a solution but it shows where the exception is raised, you can see we enter this block after verifying a `happi` import but without checking if there is a config.

Solving this looks like a pretty annoying refactor.

## Steps to Reproduce (for bugs)

1. Unset any `happi` config environment variables
2. Make sure `happi` is still installed
3. Invoke `typhos` with an explicit class and args

## Context

I was just trying to help review the sick composite signal panel PR

## Your Environment

Current master typhos in development mode

ZLLentz:
I was going to hash out a fix for this today but I got side-tracked

klauer:
Closed by #281. If I missed something, please reopen this.

PR #281, opened 2020-04-28
FIX: do not require happi on the command line

klauer:
A quick attempt at closing #271, allowing me to run `typhos` with full `class[{kwargs}]` syntax without having to jump through hoops and specify a temporary HAPPI_CFG.

Not sure if @ZLLentz saw more here than I did. Perhaps it could be handled better?

### DIFF
--- a/typhos/cli.py
+++ b/typhos/cli.py
@@ -99,7 +99,7 @@ def create_suite(devices, cfg=None, fake_devices=False):
     try:
         happi_client = _create_happi_client(cfg)
     except Exception:
-        logger.exception("Unable to create a happi client.")
+        logger.debug("Unable to create a happi client.")
         happi_client = None
 
     # Load and add each device
--- a/typhos/cli.py
+++ b/typhos/cli.py
@@ -11,7 +11,7 @@ from qtpy.QtWidgets import QApplication, QMainWindow
 
 import pcdsutils
 import typhos
-from ophyd.sim import make_fake_device, clear_fake_device
+from ophyd.sim import clear_fake_device, make_fake_device
 
 logger = logging.getLogger(__name__)
 app = None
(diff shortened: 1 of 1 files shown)


typhos/cli.py
```diff
@@ -79,24 +79,28 @@ def typhos_cli_setup(args):
...
-    if happi_enabled:
-        if typhos.plugins.happi.HappiClientState.client:
-            logger.debug("Using happi Client already registered with Typhos")
-            client = typhos.plugins.happi.HappiClientState.client
-        else:
-            logger.debug("Creating new happi Client from configuration")
-            client = happi.Client.from_config(cfg=cfg)
+        happi_client = _create_happi_client(cfg)
+    except Exception:
+        logger.exception("Unable to create a happi client.")
```
hhslepicka, line 102:
```suggestion
        logger.debug("Unable to create a happi client.")
```

typhos/cli.py
```diff
@@ -79,24 +79,28 @@ def typhos_cli_setup(args):
...
-    if happi_enabled:
-        if typhos.plugins.happi.HappiClientState.client:
-            logger.debug("Using happi Client already registered with Typhos")
-            client = typhos.plugins.happi.HappiClientState.client
-        else:
-            logger.debug("Creating new happi Client from configuration")
-            client = happi.Client.from_config(cfg=cfg)
+        happi_client = _create_happi_client(cfg)
+    except Exception:
+        logger.debug("Unable to create a happi client.")
```
ZLLentz, line 102: One thought: this has the potential to hide obscure misconfiguration bugs, and drops any known exceptions and their accompanying error messages from `happi.Client.from_config`. The output is a bit messy but I suggest including the `exc_info` in the debug statement here.
```suggestion
        logger.debug("Unable to create a happi client.", exc_info=True)
```

PR #289, opened 2020-05-04
FIX: Prime signal value before describe.

### LINKED ISSUE #288: Regression with ImageButton
## Expected Behavior
2D array data should show Image Button.

## Current Behavior
It is showing a text edit and a label with the array in it.

1. Open a typhos display with a camera
2. Enable the Omitted Signals
3. Look for the shaped_image signal and note the entry.

## Context
This seems to be a regression caused by #282 . I need to investigate further.

hhslepicka:
Maybe not the best solution but a solution so far...
Closes #288 

## Description
While debugging this issue I noticed that if a value never arrived and describe is called at a signal it may report the wrong information which is what happened with the Image Button.

This fix calls get to force a prime of the value at the signal. I am not sure yet if there is a better way to address this issue. I will look into it for a bit to try and identify if the value callback was executed at least once so we can avoid this "HACK".

## Motivation and Context
Fix the issue #288.

## How Has This Been Tested?
Locally tested with IOC.

N/A

Before the fix:
image

After proposed fix:
image

### DIFF
--- a/typhos/cache.py
+++ b/typhos/cache.py
@@ -73,6 +73,8 @@ class _GlobalDescribeCache(QtCore.QObject):
             # Force initial value readout otherwise _readback was never
             # set and that causes issues with more complex describe
             # types such as DerivedSignal and NDDerivedSignal.
+            # TODO: This is a temporary fix and must be removed once
+            #       https://github.com/bluesky/ophyd/pull/858 is merged
             obj.get()
             return obj.describe()[obj.name]
         except Exception:
--- a/typhos/cache.py
+++ b/typhos/cache.py
@@ -70,6 +70,10 @@ class _GlobalDescribeCache(QtCore.QObject):
     def _describe(self, obj):
         """Retrieve the description of ``obj``."""
         try:
+            # Force initial value readout otherwise _readback was never
+            # set and that causes issues with more complex describe
+            # types such as DerivedSignal and NDDerivedSignal.
+            obj.get()
             return obj.describe()[obj.name]
         except Exception:
             logger.error("Unable to connect to %r during widget creation",
--- a/typhos/cache.py
+++ b/typhos/cache.py
@@ -1,5 +1,6 @@
 import functools
 import logging
+import ophyd
 
 from qtpy import QtCore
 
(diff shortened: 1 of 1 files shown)


klauer:
Looks like we need to fix this in ophyd. A shame that the minor release was just done last Friday!

hhslepicka:
That is true, @klauer as I see the same issue when I run just the ophyd device at the python interpreter and call describe.

ZLLentz:
This seems like a good fix to me.

What would the corresponding fix be on the `ophyd` side? Just wrapping a `get` call directly in `describe` if we never got data?

klauer:
@ZLLentz yeah, `describe` should handle calling `get()` itself.

The reason we don't want this on the typhos side is that it could cause up to a 2x performance decrease, that is `.get()` being called prior to describe and during describe.

ZLLentz:
Yeah, we'd have to essentially do exactly what `ophyd` should be doing: only performing the `get` if it is absolutely needed for a successful `describe`

ZLLentz:
But a quick skim of `ophyd` shows that it appears to be doing this on all the `Signal` base classes. Which one is doing the wrong thing?

klauer:
@hhslepicka is digging into the details, but I'd assume it's either https://github.com/bluesky/ophyd/blob/master/ophyd/areadetector/base.py#L25 or the underlying `DerivedSignal`

hhslepicka:
That is correct... I am looking at the DerivedSignal describe.

hhslepicka:
PR with an Ophyd Fix was opened: https://github.com/bluesky/ophyd/pull/858

hhslepicka:
I added a note into the code for removal of this additional `.get` once the Ophyd PR is merged.
What do you think about merging this in and opening an issue related to the Ophyd PR for reverting the changes here? @klauer & @ZLLentz

ZLLentz:
Any worry about the overhead of calling `get` on every PV? Does this ever use a cached value or does it always fetch a new one?

hhslepicka:
The worry is a constant in my life... I will check on the cache argument and I will be right back with an answer for that.

hhslepicka:
Ok... after a quick look at the code the answer gray...
If a signal uses monitor it will return the monitored value. If it does not use monitor it will hit the control system. If it is a soft signal or derived it will return the last value received by the signal.

This fix is far from being a good one in terms of performance... other options would include probably patching the Ophyd class itself via one of our packages but that looks too invasive.

Another idea here:
This issue seems to affect only NDDerivedSignals... I can put a check wrapping this obj.get.

klauer:
A temporary check for `NDDerivedSignal` + a `get()` call that we can remove down the line doesn't seem so bad.

hhslepicka:
Change added... now it works and only calls get for NDDerivedSignals.
Sorry that it took me all that to make this little change.

hhslepicka:
Restarting CI as it seems like I don't have permission at Travis-CI to rerun jobs.

Issue #285, opened 2020-05-01
Signal describe cache persistence


klauer:
With #282, a cache is used for holding signal information.
This cache starts with a clean slate for each typhos process.

It may be wise to persist the cache and/or make it accessible to multiple processes.

Potential options may include:
* An sqlite database (may allow for concurrent access, though concerns of using sqlite over NFS due to locking.)
* a `PersistentDict`/`zict` (as in bluesky/dask).

hhslepicka:
Would a microservice help here?

klauer:
Any specific microservice you're thinking of @hhslepicka?

hhslepicka:
A simple tornado or flask app for the facility serving the describe data maybe? That would also allow us to start measuring which devices are most used and etc in the future...
Maybe I am creating a beast that does not need to exist

klauer:
I think eventually that's probably a good idea, so long as we can keep:

The time to get the describe cache `<<` the time it takes to connect and call `.describe()`

Or perhaps `<` / on-the-order-of may be sufficient, depending on how useful the offline features become. That is, creating the right widgets even when the IOC is stopped, for example.

hhslepicka:
Reevaluating my statement (which now thinking more about it, it was dumb)... there will be no service faster than direct access to the cache file (unless storage is really 💩 )

klauer:
If we ever want concurrency above the per-user level (e.g., beamline-wide or facility-wide), we'll need some sort of service for sure.

It's not impossible that NFS may be so slow as to make even #290 not a viable solution 🤷

klauer:
As noted in the closure of #290, file-based solutions just don't work for us.

I think a centralized database may be our best option here.

Wishlist
* Entirely optional
* Easy-to-configure - maybe with a pre-configured docker image even
* Faster than `obj.describe()`

Should be easy!

PR #290, opened 2020-05-04
User config-stored sqlite `obj.describe()` cache

klauer:
## Description
* Adds `utils.get_config_path` for getting a valid configuration path for the 3 main platforms
    - This can be overridden by the environment variable `TYPHOS_CONFIG_PATH`
* Adds `cache._DescribeDatabase`, an sqlite3-backed database for storing ophyd object descriptions.
    - Path defaults to `$TYPHOS_CONFIG_PATH/typhos_describe.sqlite`
    - This can be overridden by env variable `$TYPHOS_DATABASE_PATH`
    - `TYPHOS_DATABASE_PATH` may be `:memory:`, should issues arise
* `cache._DescribeDatabase` and `cache._GlobalDescribeCache` are loosely tied together
   - A single Qt signal connects the in-memory cache to the persistent cache
   - A single attribute in the in-memory cache queries the persistent cache

## Motivation and Context
With a persistent cache, the backing control system does not need to be running for the correct widgets to be created with typhos.

First pass at (but does not close) #285 

## How Has This Been Tested?
Locally and with a couple unit tests, provided.

Note: *This _must_ be tested for viability on NFS prior to potential merging.*

Docstrings, so far

## TODO
- [ ] Do not cache fake signals
- [x] More explicit option to disable this caching, or is `:memory:` above sufficient?

### DIFF
--- a/typhos/cache.py
+++ b/typhos/cache.py
@@ -35,7 +35,9 @@ def get_global_describe_cache():
         db_path = get_describe_database_path()
         persistent_cache = _DescribeDatabase() if db_path else None
         _GLOBAL_DESCRIBE_CACHE = _GlobalDescribeCache(persistent_cache)
-        persistent_cache.describe_cache = _GLOBAL_DESCRIBE_CACHE
+        if persistent_cache is not None:
+            # Hook the persistent cache up to the global describe cache:
+            persistent_cache.describe_cache = _GLOBAL_DESCRIBE_CACHE
 
     return _GLOBAL_DESCRIBE_CACHE
 
--- a/typhos/cache.py
+++ b/typhos/cache.py
@@ -516,7 +516,7 @@ class _DescribeDatabase(collections.abc.Mapping):
         t0 = time.time()
         self._preload()
         elapsed = time.time() - t0
-        self.log.info('Took %d ms to preload the cache', int(elapsed * 1000))
+        self.log.debug('Took %d ms to preload the cache', int(elapsed * 1000))
 
     def _preload(self):
         """Pre-load all entries from the database."""
--- a/typhos/cache.py
+++ b/typhos/cache.py
@@ -1,4 +1,5 @@
 import ast
+import collections.abc
 import fnmatch
 import functools
 import logging
@@ -30,7 +31,13 @@ def get_global_describe_cache():
     """Get the _GlobalDescribeCache singleton."""
     global _GLOBAL_DESCRIBE_CACHE
     if _GLOBAL_DESCRIBE_CACHE is None:
-        _GLOBAL_DESCRIBE_CACHE = _GlobalDescribeCache()
+        # NOTE: it's possible to disable the cache by clearing the environment
+        # variable `TYPHOS_DATABASE_PATH`.
+        db_path = get_describe_database_path()
+        persistent_cache = _DescribeDatabase() if db_path else None
+        _GLOBAL_DESCRIBE_CACHE = _GlobalDescribeCache(persistent_cache)
+        persistent_cache.describe_cache = _GLOBAL_DESCRIBE_CACHE
+
     return _GLOBAL_DESCRIBE_CACHE
 
 
@@ -93,7 +100,7 @@ class _GlobalDescribeCache(QtCore.QObject):
 
     new_description = QtCore.Signal(object, dict)
 
-    def __init__(self):
+    def __init__(self, persistent_cache=None):
         super().__init__()
         self.connect_thread = utils.ObjectConnectionMonitorThread(parent=self)
         self.connect_thread.connection_update.connect(self._connection_update)
@@ -101,7 +108,7 @@ class _GlobalDescribeCache(QtCore.QObject):
 
         self._in_process = set()
         self.cache = {}
-        self.persistent_cache = {}
+        self.persistent_cache = persistent_cache or {}
 
     def clear(self):
         """Clear the cache."""
@@ -378,7 +385,7 @@ class _GlobalDisplayPathCache:
             self.paths.append(path)
 
 
-class _DescribeDatabase(QtCore.QObject):
+class _DescribeDatabase(collections.abc.Mapping):
     """
     An Sqlite3 database-backed cache of ophyd object descriptions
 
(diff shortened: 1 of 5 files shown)


klauer:
To reiterate from the PR description, this needs testing with real devices and real machines (e.g., not personal laptops where the sqlite file is on a fast SSD).

ZLLentz:
I'll give this a whirl on NFS land

typhos/utils.py
```diff
@@ -1058,3 +1069,30 @@ def get_text(item):
...
+    if platform.system().lower() == 'windows':
+        path = os.getenv('APPDATA') or '~'
+    elif platform.system().lower() == 'darwin':
+        path = pathlib.Path('~') / 'Library' / 'Caches'
+    else:
+        path = os.getenv('XDG_CACHE_HOME') or '~/.cache'
+
+    # Allow TYPHOS_CONFIG_PATH to override the above operating system-specific
+    # entries:
+    path = os.environ.get('TYPHOS_CONFIG_PATH', path / 'typhos')
```
ZLLentz, line 1085: This fails on linux because path is a string, unlike macos which is pre-converted to a `pathlib.Path`

ZLLentz:
On psbuild-rhel7-01 using the default cache location and the latest tags of everything else:

It took about 1 minute from first log message to last while writing the cache, though the screen wasn't responsive to resizing/scrolling/closing after this for a while. It also wouldn't close very quickly... I suspect something was still initializing. This is a slowdown from the previous 40 seconds, but this is expected because we're checking and writing to a cache.

On second load (reading the cache) it took about 2 minutes 40 seconds but was very responsive immediately. This suggests writing a cache is counterproductive on NFS since we used to be able to do this in 1 minute.

Third load was the same as the second load.

A quick test in IPython after the cache had been written out using
 `typhos "pcdsdevices.device_types.PPM[{'prefix':'IM3L0:PPM', 'name':'im3l0'}]"`:

```
from typhos.cache import _DescribeDatabase
from pcdsdevices.device_types import PPM

db = _DescribeDatabase()
im3l0 = PPM('IM3L0:PPM', name='im3l0')
sig = im3l0.led

In [1]: %timeit db.get(sig)
2.11 ms ± 49.6 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
In [2]: len(list(im3l0.walk_signals()))
Out[2]: 287
In [3]: def all_gets():
    ...:     for walk in im3l0.walk_signals():
    ...:         db.get(walk.item)
    ...:
In [4]: %timeit all_gets()
620 ms ± 16.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This suggests that the database action is fine and should only amount to a little over half a second of time spent across the full device... meaning that NFS seems ok and there is some other problem?

I have not had a chance to dive into the code yet, except the minimum required to set up the `IPython` test.

typhos/utils.py
klauer: D'oh - I had rewritten this like 3 times, thinking I had gotten the most simple version. Will fix.

klauer:
Thanks for the test effort, @ZLLentz. As-is, the database-backed cache doesn't make any effort to batch reads/writes, which could significantly improve performance over NFS. 

That is: every `describe` is immediately committed (saved) into the database, and every `get` call cache miss results in a database read.

Your near 3 minute result is very confusing. I think I'll grab your database and attempt to reproduce it...

klauer:
Force-pushed after rebasing on conflicting cache changes.

typhos/cache.py
```diff
@@ -3,7 +3,6 @@
 import fnmatch
 import functools
 import logging
-import ophyd
```
hhslepicka: it was me! :D my bad!

klauer: Even worse, I pushed a commit that shouldn't be here:  52e2f7c ; re-force pushing to save face

klauer:
Added, then removed, queueing signal work by @hhslepicka which looks promising but will require fixing up a bunch of tests; moving to a separate PR.

klauer:
(re-requesting reviews from @hhslepicka and @ZLLentz ) 
I think this is probably good enough for now. Most of the code here is reasonably well isolated - as in it does not touch other parts of the codebase.

If there's even a bit of concern, we could perhaps make the default setting to be 'no persistent database unless specified by the environment'.

hhslepicka:
The cache code here is solid and I think that it is a great addition to the codebase.
I would vote to merge it in so we can address the remaining parts at separated PRs.

ZLLentz:
I'll clear my cache and give this a try again

ZLLentz:
This still is very slow on the NFS with the PPM, slower than without the patch (2x slowdown, from 1 minute to 2 minutes). Loading from an empty cache and from a full cache both took around 2 minutes.

typhos/cache.py
```diff
@@ -341,3 +386,291 @@ def add_path(self, path):
...
+        # Support __getitem__ in row access:
+        self._con.row_factory = sqlite3.Row
+
+        self.log.info('Describe database path: %s', path)
+        self._con.execute(self._create_query)
+
+        t0 = time.time()
+        self._preload()
+        elapsed = time.time() - t0
+        self.log.info('Took %d ms to preload the cache', int(elapsed * 1000))
```
ZLLentz, line 519:
```suggestion
        self.log.debug('Took %d ms to preload the cache', int(elapsed * 1000))
```
This isn't useful information for the end-user so it belongs in debug:
```
[2020-05-07 14:42:13] - INFO - Loading "pcdsdevices.device_types.PPM[{'prefix':'IM3L0:PPM'}]" ...
[2020-05-07 14:42:15] - INFO - Loading Tools ...
[2020-05-07 14:42:15] - INFO - Describe database path: /reg/neh/home4/zlentz/.cache/typhos/typhos_describe.sqlite
[2020-05-07 14:42:15] - INFO - Took 2 ms to preload the cache
[2020-05-07 14:42:16] - INFO - Adding devices ...
[2020-05-07 14:42:42] - INFO - Launching application ...
[2020-05-07 14:44:02] - INFO - Execution complete!
```
```
[2020-05-07 14:44:20] - INFO - Loading "pcdsdevices.device_types.PPM[{'prefix':'IM3L0:PPM'}]" ...
[2020-05-07 14:44:21] - INFO - Loading Tools ...
[2020-05-07 14:44:22] - INFO - Describe database path: /reg/neh/home4/zlentz/.cache/typhos/typhos_describe.sqlite
[2020-05-07 14:44:28] - INFO - Took 6102 ms to preload the cache
[2020-05-07 14:44:28] - INFO - Adding devices ...
[2020-05-07 14:46:01] - INFO - Launching application ...
[2020-05-07 14:46:17] - INFO - Execution complete!
```

klauer: Agreed and accepted.

However, that 6 seconds is atrocious. Perhaps that indicates this just shouldn't be merged at all (or at minimum an alternative pre-loading method should be done)

klauer:
Yeah, I think this isn't worth merging given the NFS issues. I don't think we'll be able to get a file-based solution on our production machines due to this.

Any alternative I can think of comes with its own set of issues:
* Use a copy of the database and work on it in-memory
   - Fast startup; but commits slow + synchronization issues
* Use a non-NFS location (then it's not available everywhere to the user)

I'm going to take the tests out of this PR and close it, as at least they are useful.

PR #291, opened 2020-05-07
ENH: Benchmarking Tools

ZLLentz:
## Description

Adds a `benchmark` submodule and some `cli` tools to help us keep track of `typhos`'s performance.  This also required a refactor of the `cli` submodule to pull apart the large functions there.

Full feature checklist:

- [x] `line_profiler` integration via cli to debug arbitrary screens interactively
- [x] Test class generator
- [x] Benchmark test cases with `caproto` and with soft devices
- [x] Benchmarking via `cli` using `line_profiler` for debugging
- [x] Benchmarking via `pytest-benchmark` for tracking performance

Some specific things to improve:

- [x] Expand test class generator for nested classes
- [x] Expand test cases to cover more ground
- [x] Clean up the argument name inconsistency mess I left in `cli.py` (and other things there)

Remaining work before merging:

- [x] Make the benchmark tests not hang afterwards
- [x] Add basic regression testing for the benchmark utilities

Stuff that I wanted to do but will not get to in this PR:

- [ ] Sort the `line_profiler` formatted output
- [ ] Allow filtering `line_profiler` output down to the function level
- [ ] Provide the raw `line_profiler` output file

## Motivation and Context

This was brought about because some of the larger screens have been taking a long time to load.

## How Has This Been Tested?

I've been testing locally, but this will be implicitly tested when I add the benchmark tests to the test suite.

I have not yet documented anything beyond docstrings.

### DIFF
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -3,7 +3,9 @@ Run the benchmark test cases using pytest-benchmark
 """
 import pytest
 
+import typhos.benchmark.utils as utils
 from typhos.benchmark.cases import unit_tests
+from typhos.benchmark.profile import profiler_context
 from typhos.suite import TyphosSuite
 
 from .conftest import save_image
@@ -11,7 +13,7 @@ from .conftest import save_image
 
 # Name the test cases using the keys, run using the values
 @pytest.mark.parametrize('unit_test_name', unit_tests.keys())
-def test_benchmark(unit_test_name, qtbot, benchmark):
+def test_benchmark(unit_test_name, qapp, qtbot, benchmark):
     """
     Run all registered benchmarks.
 
@@ -27,3 +29,11 @@ def inner_benchmark(unit_test, qtbot):
         qtbot.add_widget(suite)
         qtbot.wait_active(suite)
     return suite
+
+
+def test_profiler(capsys):
+    """Super basic test that hits most functions here"""
+    with profiler_context(['typhos.benchmark.utils']):
+        utils.get_native_functions(utils)
+    output = capsys.readouterr()
+    assert 'get_native_functions' in output.out
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -5,6 +5,8 @@ import pytest
 
 from typhos.benchmark.cases import benchmark_tests
 
+from .conftest import save_image
+
 
 # Name the test cases using the keys, run using the values
 @pytest.mark.parametrize('test_function', benchmark_tests.keys())
@@ -14,4 +16,5 @@ def test_benchmark(test_function, benchmark):
 
     These typically just open and close a particular typhos screen.
     """
-    benchmark(benchmark_tests[test_function])
+    window = benchmark(benchmark_tests[test_function])
+    save_image(window, 'test_benchmark_' + test_function)
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -1,0 +1,17 @@
+"""
+Run the benchmark test cases using pytest-benchmark
+"""
+import pytest
+
+from typhos.benchmark.cases import benchmark_registry
+
+
+# Name the test cases using the keys, run using the values
+@pytest.mark.parametrize('test_function', benchmark_registry.keys())
+def test_benchmark(test_function, benchmark):
+    """
+    Run all registered benchmarks.
+
+    These typically just open and close a particular typhos screen.
+    """
+    benchmark(benchmark_registry[test_function])
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -3,6 +3,7 @@ Run the benchmark test cases using pytest-benchmark
 """
 import pytest
 
+from epics import PV
 import typhos.benchmark.utils as utils
 from typhos.benchmark.cases import unit_tests
 from typhos.benchmark.profile import profiler_context
@@ -12,12 +13,14 @@ from .conftest import save_image
 
 # Name the test cases using the keys, run using the values
 @pytest.mark.parametrize('unit_test_name', unit_tests.keys())
-def test_benchmark(unit_test_name, qapp, qtbot, benchmark):
+def test_benchmark(unit_test_name, qapp, qtbot, benchmark, monkeypatch):
     """
     Run all registered benchmarks.
 
     These typically just open and close a particular typhos screen.
     """
+    # Crudely permenant patch here to get around cleanup bug
+    PV.count = property(lambda self: 1)
     suite = benchmark(inner_benchmark, unit_tests[unit_test_name], qtbot)
     save_image(suite, 'test_benchmark_' + unit_test_name)
 
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -3,11 +3,11 @@ Run the benchmark test cases using pytest-benchmark
 """
 import pytest
 
-from typhos.benchmark.cases import benchmark_registry
+from typhos.benchmark.cases import benchmark_tests
 
 
 # Name the test cases using the keys, run using the values
-@pytest.mark.parametrize('test_function', benchmark_registry.keys())
+@pytest.mark.parametrize('test_function', benchmark_tests.keys())
 def test_benchmark(test_function, benchmark):
     """
     Run all registered benchmarks.
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -14,4 +14,4 @@ def test_benchmark(test_function, benchmark):
 
     These typically just open and close a particular typhos screen.
     """
-    benchmark(benchmark_registry[test_function])
+    benchmark(benchmark_tests[test_function])
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -6,7 +6,6 @@ import pytest
 import typhos.benchmark.utils as utils
 from typhos.benchmark.cases import unit_tests
 from typhos.benchmark.profile import profiler_context
-from typhos.suite import TyphosSuite
 
 from .conftest import save_image
 
--- a/tests/test_benchmark.py
+++ b/tests/test_benchmark.py
@@ -3,18 +3,27 @@ Run the benchmark test cases using pytest-benchmark
 """
 import pytest
 
-from typhos.benchmark.cases import benchmark_tests
+from typhos.benchmark.cases import unit_tests
+from typhos.suite import TyphosSuite
 
 from .conftest import save_image
 
 
 # Name the test cases using the keys, run using the values
-@pytest.mark.parametrize('test_function', benchmark_tests.keys())
-def test_benchmark(test_function, benchmark):
+@pytest.mark.parametrize('unit_test_name', unit_tests.keys())
+def test_benchmark(unit_test_name, qtbot, benchmark):
     """
     Run all registered benchmarks.
 
     These typically just open and close a particular typhos screen.
     """
-    window = benchmark(benchmark_tests[test_function])
-    save_image(window, 'test_benchmark_' + test_function)
+    suite = benchmark(inner_benchmark, unit_tests[unit_test_name], qtbot)
+    save_image(suite, 'test_benchmark_' + unit_test_name)
+
+
+def inner_benchmark(unit_test, qtbot):
+    suite, context = unit_test()
+    with context:
+        qtbot.add_widget(suite)
+        qtbot.wait_active(suite)
+    return suite
--- a/typhos/app.py
+++ b/typhos/app.py
@@ -1,0 +1,36 @@
+"""This module defines methods for launching full typhos applications."""
+import logging
+
+from qtpy.QtWidgets import QApplication, QMainWindow
+
+logger = logging.getLogger(__name__)
+qapp = None
+
+def get_qapp():
+    """Returns the global QApplication, creating it if necessary."""
+    global qapp
+    if qapp is None:
+        logger.debug("Creating QApplication ...")
+        qapp = QApplication([])
+    return qapp
+
+
+def launch_suite(suite):
+    """Creates a main window and execs the application."""
+    window = QMainWindow()
+    window.setCentralWidget(suite)
+    window.show()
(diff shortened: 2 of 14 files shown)


typhos/benchmark/cases.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    proc = Process(target=run_caproto_ioc, args=(device_class, prefix))
+    proc.start()
+    yield
+    proc.terminate()
+
+
+def test_flat_soft():
+    """Launch typhos using a flat device with no EPICS connections."""
+    launch_from_devices([FlatSoft('TEST:', name='test')])
```
klauer, line 66: Assuming one IOC per test, we found in caproto that a random prefix helps ensure that tests won't interfere with one another:

https://github.com/caproto/caproto/blob/96c488538b23757406ea6af3846fb09883093785/caproto/tests/conftest.py#L168

typhos/benchmark/cases.py
```diff
@@ -0,0 +1,77 @@
...
+                                  include_prefix=False, num_signals=100)
+FlatEpics = make_test_device_class(name='FlatEpics', signal_class=EpicsSignal,
+                                   include_prefix=True, num_signals=100)
+
+
+def run_benchmarks(benchmarks):
+    raise NotImplementedError('TODO: this')
+
+
+def run_caproto_ioc(device_class, prefix):
```
klauer, line 28: This is cool 👍

typhos/benchmark/device.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def make_test_device_class(name='TestClass', signal_class=Signal,
+                           include_prefix=False, num_signals=10):
+    """
+    Creates a test Device class.
+
+    Parameters
+    ----------
+    name: str, optional
```
klauer, line 23: Docstring conventions as per new style guide

```suggestion
    name : str, optional
```

typhos/benchmark/device.py
```diff
@@ -0,0 +1,50 @@
...
+        The name of the class.
+        Defaults to 'TestClass'.
+
+    signal_class: type, optional
+        Picks which type of signal to use in the Device.
+        Defaults to ophyd.signal.Signal.
```
klauer, line 29:
```suggestion
        Defaults to :class:`ophyd.Signal`.
```

typhos/benchmark/profile.py
```diff
@@ -0,0 +1,120 @@
...
+    """Saves the formatted profiling results to filename."""
+    profiler = get_profiler()
+    with open(filename, 'w') as fd:
+        profiler.print_stats(fd)
+
+
+def print_results():
+    """Prints the formatted results directly to screen."""
+    profiler = get_profiler()
+    profiler.print_stats()
```
klauer, line 63: My preferences for the output, but maybe config options would be nice.

* Remove items not called and switch to milliseconds as default units:
```suggestion
    profiler.print_stats(stripzeros=True, output_unit=1e-3)
```

typhos/benchmark/profile.py
```diff
@@ -0,0 +1,120 @@
...
+    return native_methods
+
+
+def get_submodules(module_name):
+    """Returns a list of the imported module plus all submodules."""
+    submodule_names = get_submodule_names(module_name)
+    return import_modules(submodule_names)
+
+
+def get_submodule_names(module_name):
```
klauer, line 95: These are all handy methods 👍

typhos/cli.py
```diff
@@ -154,43 +179,92 @@ def create_suite(devices, cfg=None, fake_devices=False):
...
-                suite.add_device(device)
-                suite.show_subdisplay(device)
-            except Exception:
-                logger.exception("Unable to add %r to TyphosSuite",
-                                 device.name)
-        return suite
+    return loaded_devs
+
+
+def suite_from_devices(devices):
```
klauer, line 185: `@classmethod` on `TyphosSuite` perhaps?

typhos/benchmark/cases.py
ZLLentz: This makes a lot of sense! I'll lift this wholesale.

typhos/benchmark/device.py
ZLLentz: Ah shoot, this habit is one that's ingrained pretty deeply for me. I'll go through and fix it everywhere.

typhos/benchmark/device.py
ZLLentz: What effect does this have on the generated docs?

typhos/benchmark/profile.py
ZLLentz: I think these make a lot of sense, I'll throw it in here and for the file dump

typhos/cli.py
ZLLentz: I think you're correct. I'll keep this conversation unresolved until I go back to clean up this file.

typhos/benchmark/device.py
klauer: I think with the new configuration Tyler's working on, we'll be able to cross-link into ophyd documentation.

typhos/benchmark/cases.py
```diff
@@ -0,0 +1,105 @@
...
+A collection of benchmarks to run for typhos.
+
+These are included as standalone functions to make it easy to pass them into
+arbitrary profiling modules.
+"""
+from contextlib import contextmanager
+from multiprocessing import Process
+import uuid
+
+from caproto.server import pvproperty, PVGroup, run
```
klauer, line 11: Add `caproto` to dev requirements?

typhos/cli.py
```diff
@@ -96,7 +114,15 @@ def _create_happi_client(cfg):
 
 
 def create_suite(devices, cfg=None, fake_devices=False):
-    """Create a TyphosSuite from a list of device names"""
+    """Create a TyphosSuite from a list of device names."""
+    if devices:
```
klauer, line 118: Fails when using `typhos` with no devices (unbound loaded_devs on line 120)

typhos/benchmark/cases.py
ZLLentz: Yes, I definitely need to do this.

typhos/cli.py
ZLLentz: Oops! Guess I broke it.

ZLLentz:
This should be getting close now. I'm not 100% sure that the method I picked to auto-close the windows during benchmarking works with the queued connections... it might be closing them early, but I can't really tell? The intention was to close the window as soon as the user would gain control.

I think this just needs some unit tests for some of the miscellaneous functions I slapped into the benchmark module. I'll send review requests on Monday because doing it on a Friday is just asking to be forgotten.

I also may remove the benchmarks from the Travis or make them smaller. Currently I'm doing 1000 signals and it takes forever to run... that might indicate a performance problem or it might be fighting against the limits of the machine I'm running on :shrug:

klauer:
Would you consider adding a (perhaps optional?) `conftest.save_image()` call at the end of every benchmark?

ZLLentz:
I added a (non-optional) save image like all the other tests have. I think we can add a command-line arg later to make all image saving optional. The images are not useful here because the screen opens up really small...

image

hhslepicka:
The screens opening in a small size is a side effect of my performance fix. We can force them to be bigger with a resize.

ZLLentz:
Remaining work here (modulo stuff found in review) is to refactor so that the `pytest-benchmark` can be done with `qtbot` and to add regression testing for some of the `benchmark` utilties. Other than that I don't anticipate huge changes to this PR.

ZLLentz:
Closing and re-opening because this seems to have been disconnected from the associated travis builds.

typhos/benchmark/profile.py
```diff
@@ -0,0 +1,94 @@
...
+    yield
+    toggle_profiler(False)
+
+    if filename is None:
+        print_results()
+    else:
+        save_results(filename)
+
+
+def setup_profiler(module_names=['typhos']):
```
klauer, line 53: Mutable in kwarg list. Should be refactored.

typhos/benchmark/profile.py
```diff
@@ -0,0 +1,94 @@
...
+    global profiler
+    if not has_line_profiler:
+        raise ImportError(_optional_err)
+    elif profiler is None:
+        profiler = LineProfiler()
+    return profiler
+
+
+@contextmanager
+def profiler_context(module_names=[], filename=None):
```
klauer, line 36: Mutable in kwarg list. Should be refactored.

typhos/benchmark/device.py
```diff
@@ -0,0 +1,81 @@
...
+
+    SignalHolder = type('SignalHolder', (Device,), signals)
+
+    if all((subdevice_layers > 0, subdevice_spread > 0)):
+        PrevClass = SignalHolder
+        while subdevice_layers > 0:
+            subdevices = {}
+            for ndev in range(subdevice_spread):
+                subdevices[f'devnum{ndev}'] = Cpt(PrevClass, f'PREFIX{ndev}:')
+            ThisClass = type(f'Layer{subdevice_layers}', (Device,), subdevices)
```
klauer, line 75: https://github.com/bluesky/ophyd/blob/07d2662b7d04879735c62909a905aa4067885216/ophyd/device.py#L1505 exists to make the intent of `type(..., (Device, )` more obvious.

typhos/app.py
```diff
@@ -0,0 +1,40 @@
...
+    return qapp
+
+
+def launch_suite(suite):
+    """Creates a main window and execs the application."""
+    window = QMainWindow()
+    window.setCentralWidget(suite)
+    window.show()
+    logger.info("Launching application ...")
+    QApplication.instance().exec_()
```
klauer, line 28:
```suggestion
    get_qapp().exec_()
```

typhos/app.py
```diff
@@ -0,0 +1,40 @@
...
+from .suite import TyphosSuite
+
+logger = logging.getLogger(__name__)
+qapp = None
+
+
+def get_qapp():
+    """Returns the global QApplication, creating it if necessary."""
+    global qapp
+    if qapp is None:
```
klauer, line 16:
```python
if QApplication.instance() is not None:
    qapp = QApplication.instance()
```

typhos/app.py
```diff
@@ -0,0 +1,40 @@
...
+    return window
+
+
+def launch_from_devices(devices, auto_exit=False):
+    """Alternate entry point for non-cli testing of loader."""
+    app = get_qapp()
+    suite = TyphosSuite.from_devices(devices)
+    if auto_exit:
+        timer = QTimer(suite)
+        timer.singleShot(0, app.exit)
```
hhslepicka, line 39: singleShot is static.
```suggestion
        QTimer.singleShot(0, app.exit)
```

typhos/suite.py
```diff
@@ -381,19 +381,52 @@ def from_device(cls, device, parent=None, tools=dict(), pin=False,
         kwargs:
             Passed to :meth:`TyphosSuite.add_device`
         """
-        display = cls(parent=parent, pin=pin)
+        return cls.from_devices([device], parent=parent, tools=tools, pin=pin,
+                                **kwargs)
+
+    @classmethod
+    def from_devices(cls, devices, parent=None, tools=dict(), pin=False,
```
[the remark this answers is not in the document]
ZLLentz: I'll switch to some benign sentinel value (since for some reason the legacy API wants `None` to mean no tools)

typhos/benchmark/utils.py
```diff
@@ -0,0 +1,159 @@
...
+    """Returns True if obj was defined in module."""
+    return module.__name__ in obj.__module__
+
+
+def get_native_functions(module):
+    """Returns all functions and methods defined in module."""
+    return get_native_methods(module, module)
+
+
+def get_native_methods(cls, module):
```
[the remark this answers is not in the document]
ZLLentz: Basically, a loop if two classes share mutual references to each other? (Or a similar less direct loop?) Seems common enough to be an issue.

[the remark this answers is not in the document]
klauer: Yeah, some cyclical reference - at any depth. Try the original method on `caproto` and that should trigger it.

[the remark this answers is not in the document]
ZLLentz: And then I assume the `native_methods` passing is just for a speedup so we create fewer lists... there's also an issue I see now where we might recurse through way too many objects. I'll also edit this to call `is_native` sooner so we don't end up in some other module and spider through someone else's package (obviously finding nothing)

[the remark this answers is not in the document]
klauer: That sounds about right. It was easier to paste my quick fixes that allowed it to work for profiling of caproto rather than make a poor attempt at explaining it 🤷‍♂️

typhos/benchmark/device.py
ZLLentz: I had no idea this was a thing, will incorporate

typhos/app.py
ZLLentz: The curse of moving old code around is that you draw attention to it!

ZLLentz:
The `pytest` portions hang after finishing whenever we have an EPICS signal, regardless of whether I start the `caproto` IOC or not. I'm on the fence between the options of:
1. Fixing the issue
2. Running only soft signals in the `pytest`
3. Removing the `pytest` portions entirely for simplicity

tests/test_benchmark.py
```diff
@@ -0,0 +1,29 @@
+"""
+Run the benchmark test cases using pytest-benchmark
+"""
+import pytest
+
+from typhos.benchmark.cases import unit_tests
+from typhos.suite import TyphosSuite
```
hhslepicka, line 7: Unused import
```suggestion

```

ZLLentz:
I'd like to relegate future non-bugfix work related to this to future PRs. Hopefully travis is happy today.

klauer:
Great! Let's get this in and get the docs PR rebased on top.

Local testing shows that this works well - we can increase coverage later.
I'm also going to open a separate PR to un-skip the cache tests. We really need to get to the bottom of that.

PR #294, opened 2020-05-07
FIX: do not fail if `unsubscribe` fails; mark nested devices

klauer:
Pulled from #290


klauer:
The `nested` attribute is mainly for information/debugging purposes now.
In the future, it could be a `QtCore.Property` which has an effect on how the display is styled.

PR #298, opened 2020-05-13
FIX: font sizes in pixels, window title, and OSX toolbar

### LINKED ISSUE #260: Switch from Point sized to Pixel sized for Fonts

klauer:
## Description
* Changes row labels and titles to be in terms of pixel units. See screenshots below and related issue for details.
* Uses stylesheet instead of code for titles
* Add title for Typhos suite to get rid of ugly `python` title
* Fixes toolbar rendering on OSX (at least for now)
* Adds `SignalPanelRowLabel` for ease of styling

Closes #260 

Before (OSX/Win32):
image

After (OSX/Win32):
image

### DIFF
--- a/typhos/suite.py
+++ b/typhos/suite.py
@@ -93,12 +93,18 @@ class TyphosSuite(TyphosBase):
     ----------
     parent : QWidget, optional
     """
+
+    DEFAULT_TITLE = 'Typhos Suite'
+    DEFAULT_TITLE_DEVICE = 'Typhos Suite - {device.name}'
+
     default_tools = {'Log': TyphosLogDisplay,
                      'StripTool': TyphosTimePlot,
                      'Console': TyphosConsole}
 
     def __init__(self, parent=None, *, pin=False):
         super().__init__(parent=parent)
+
+        self._update_title()
 
         self._tree = ParameterTree(parent=self, showHeader=False)
         self._tree.setAlternatingRowColors(False)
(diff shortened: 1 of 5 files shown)


klauer:
Linux screenshots for comparison's sake would be appreciated, if possible.

hhslepicka:
Before (Linux - fastx3 to tid-pc94331)
image (4)

After (Linux - fastx3 to tid-pc94331)
Screen Shot 2020-05-13 at 11 20 06 AM

klauer:
Looks pretty good - thanks @hhslepicka.

typhos/suite.py
```diff
@@ -328,6 +334,20 @@ def tools(self):
...
+
+        Parameters
+        ----------
+        device : ophyd.Device, optional
+            Device to indicate in the title.
+        """
+        title_fmt = (self.DEFAULT_TITLE if device is None
+                     else self.DEFAULT_TITLE_DEVICE)
+
+        self.setWindowTitle(title_fmt.format(self=self, device=device))
```
ZLLentz, line 349: I notice we're passing in `self` to the formatter. Is there interesting information that we have in `TyphosSuite` that could be used in a title?

klauer: Perhaps there is (or will be); not sure.

hhslepicka:
@ZLLentz , the nuance here is that systems with high DPI will properly render if sizes are in pixel and not point which is absolute value.

PR #305, opened 2020-05-13
FIX: Positioners

klauer:
## Description
* Change positioner templates to match up with class MRO search naming
   - Now anything that inherits from `PositionerBase` will have the template as an option (`EpicsMotor`, `PCDSMotorBase`, etc.)
* Make `TyphosPositionerWidget` much less magical
   - Add designable Qt Properties that allow for setting attribute names
   - If you know that "user_readback" is actually "rdbk" because of _reasons_, then set it appropriately in your template
    - Updated docstring should reflect this

* Do not call `thread.terminate()` on status threads (crashed locally), but rather disconnect
* Work with the new ophyd status API (pin changed to v1.5.0+)
* Remove `grab_kind` (implementation _may be_ flawed; no longer in use)

## How Has This Been Tested?
Existing tests are still passing locally.

Docstrings

image


ZLLentz:
Looks pretty clean, I'll see if I still have a sim twincat motor running somewhere to give it a test drive with

typhos/positioner.py
```diff
@@ -122,90 +228,183 @@ def _get_position(self):
...
+        """Link the positioner high limit switch with the ui element."""
+        if signal is None:
+            widget.hide()
+
+    @_linked_attribute('low_limit_switch_attribute', 'ui.low_limit')
+    def _link_low_limit(self, signal, widget):
+        """Link the positioner lower limit with the ui element."""
+        return signal is not None
+
+    @_linked_attribute('high_limit_switch_attribute', 'ui.high_limit')
```
ZLLentz, line 262: I think there is a typo in here and above, since the soft limit positions aren't making it into the UI (instead, the 1 or 0 for hit/not hit are used). But after swapping this in there are exceptions raised.
```suggestion
    @_linked_attribute('high_limit_attribute', 'ui.high_limit')
```

ZLLentz:
This was smooth overall, a couple observations/thoughts/ideas:
- Move timeouts. These are an error in ophyd and in the screen as well, but I'm not sure where they should be configured. It was annoying while testing because I didn't feel like I had a chance to choose (perhaps I could have passed it into the cli?) and as a result I got good old red in the terminal. The default appears to be 10 seconds. Does every `ophyd` device with a motor in it need to define a move timeout? Does every `happi` entry that is solely a motor need to define a move timeout? Does the positioner widget need a move timeout field in it?
- Errors. Should errors passed up through the Status object make it into the screen somewhere as more than just a red outline? Or it is ok to leave these in the terminal? You can trivially induce an error for testing here by moving slow enough to trigger the move timeout.

I also had a status error on hitting the bottom limit for `PLC:TST:MOT:SIM:03`, but that's probably an `ophyd` artifact of some kind.

typhos/positioner.py
klauer: Great catch. My eyes were going buggy reading the similar string/method names. Pushed a fix.

Can you give me details on the exception? Mine is working with EpicsMotor, at least.

klauer:
I like where you're going with that train of thought.

For point (1), I don't really have a good plan. Timeouts in upstream ophyd have always been a source of ~complaints~ ~disagreement~ healthy discussion. Maybe we come up with an agreeable short-term solution and then make an issue for a long-term solution discussion?

For (2) - very good point. Maybe a `raise_to_operator` dialog when anything but `WaitTimeout` is returned (the source of that 10 second delay)?

typhos/positioner.py
ZLLentz: Using `typhos "pcdsdevices.device_types.BeckhoffAxis[{'prefix': 'PLC:TST:MOT:SIM:03'}]"` on psbuil-rhel7-01 and switching to the positioner template:
```
[2020-05-14 13:32:30] - ERROR - device.low_limit => self.ui.low_limit (signal: 0.0 widget: <pydm.widgets.label.PyDMLabel object at 0x7f84380ac1f8>)
  File "/reg/neh/home/zlentz/pydev/typhos/positioner.py", line 70, in wrapped
    _link_signal_to_widget(signal, widget)
  File "/reg/neh/home/zlentz/pydev/typhos/positioner.py", line 27, in _link_signal_to_widget
    plugins.register_signal(signal)
  File "/reg/neh/home/zlentz/pydev/typhos/plugins/core.py", line 27, in register_signal
    if signal.name in signal_registry:
AttributeError: 'float' object has no attribute 'name'
```

klauer: Ah, OK, what's happening there is that it's trying to link `device.low_limit` to the UI element `.low_limit`, but `device.low_limit = float()` - likely a property?

I envision this being fixed by per-class templates which indicate the correct attributes. Depending on just how annoying that ends up being, we might circle back and be more consistent in our component names? Or `[alternate solution here]`...

klauer: https://github.com/pcdshub/pcdsdevices/blob/f153abeea529d4eaa29e5cd786a24a39dd0dc210/pcdsdevices/epics_motor.py#L65-L70

klauer:
Preview of the failure status label:
image

Tooltip gives the full text.

typhos/positioner.py
klauer: Think this is a reasonable path?

ZLLentz: Got it, so for `EpicsMotor` subclasses we'd set that property to `low_travel_limit`, for example?

ZLLentz: https://github.com/bluesky/ophyd/blob/eee8797e79448a1dc02c65952af475563f4d260a/ophyd/epics_motor.py#L67

ZLLentz: Actually, it seems to me that `low_limit` is always a property, even in `PositionerBase`: https://github.com/bluesky/ophyd/blob/eee8797e79448a1dc02c65952af475563f4d260a/ophyd/positioner.py#L143

klauer: Ohh, good! That actually helps quite a bit. I'll change the default to that `low_limit_travel` then.

So it'll use that, if available, or fall back to a static `low_limit, high_limit = obj.limits` query.

klauer: Resolved in bad30dc

klauer:
I'd like to be able to remove the 'linking' mess in `positioner.py` by way of the attribute proposal in #312. This would make custom screens working on a per-attribute/signal level easy; potentially not even requiring any python code.

PR #317, opened 2020-05-15
REF: use less magical forward_property in favor of forward_properties

klauer:
New qtpyinheritance dependency _should_ be picked up through new release pf pcdsutils, with this functionality 🤞


klauer:
Dependency is picked up and it's working 👍 

With all that magic, it only saved approximately 2 lines of code.

hhslepicka:
Still saved more than 0. 👍

ZLLentz:
You also saved 3 calls to `locals().update()`, which is a construct that makes me very sad

PR #323, opened 2020-05-15
REF: rework default templates (misc panel removal; add normal panel)

### LINKED ISSUE #319: Remove 'miscellaneous' panels
`ophyd.Kind.omitted` means "omit from read/configuration". 
Currently, these signals are added to a 'miscellaneous' tab.

As we now have easier control over hiding/showing kinds, panels by default that show signals we _probably_ don't care about no longer need to be there.

## Possible Solution
Remove miscellaneous panels from the current templates.

## Context
Lifted from a discussion in #250 between @klauer @ZLLentz and @hhslepicka

### LINKED ISSUE #320: Clean default templates, adding a 'read' (or normal) panel
Available kinds:
* Hinted (= hinted | normal)
* Normal
* Config
* Omitted

Desired layout: 

```
Title     [Switcher] [Config]
-----
[Hinted signal panel]

[ (X) Normal panel tab ] [ Configuration panel tab ]
|---------------------------------------------------|
|  Normal panel: Normal but not hinted signals      |
|  Configuration panel: config kind only            |
|---------------------------------------------------|
```

Thoughts relevant to this suggested change:

_Originally posted by @ZLLentz in https://github.com/pcdshub/typhos/issues/250#issuecomment-629479660_
> ... I like the read/config as panel tabs, but only because we may be losing the omitted tab and the config tab will be lonely. I don't suppose it would hurt to do some basic heuristics on the number of hinted + reads and create the panel tab for read only if there are a lot of them?

(_Originally posted by @hhslepicka in https://github.com/pcdshub/typhos/issues/250#issuecomment-629495529_):
> - I don't believe we will have many smaller devices based on what I saw at the pcdsdevices and looking at also devices ready to play from Ophyd.
> - Since the ratio of small devices to large ones is pending to large devices I would say it should take precedence on design.
> - Even if it becomes a bit worst for smaller devices, I believe that the trend is for them to get templates, right?

klauer:
Closes #320 
Closes #319 

Adds back some OSX fixes lost in a rebase, and removes border for the scroll area.

Previews for a BeckhoffAxis, made large enough to not show the vertical scrollbar:

Embedded positioner
-----------------------
image

Detailed screen
----------------
image

Detailed positioner
--------------------
image

Engineering
---------------

image

Tree does not change.

### DIFF
--- a/typhos/app.py
+++ b/typhos/app.py
@@ -27,6 +27,8 @@ def launch_suite(suite):
     """Creates a main window and execs the application."""
     window = QMainWindow()
     window.setCentralWidget(suite)
+    window.setWindowTitle(suite.windowTitle())
+    window.setUnifiedTitleAndToolBarOnMac(True)
     window.show()
     logger.info("Launching application ...")
     get_qapp().exec_()
--- a/tests/test_display.py
+++ b/tests/test_display.py
@@ -43,19 +43,26 @@ def test_device_display(device, motor, qtbot):
             utils.get_all_signals_from_device(device, filter_by=filter_by)
         )
 
-    def check_read_panel(device):
-        '''normal or hinted signals'''
+    def check_hint_panel(device):
         device_signals = signals_from_device(device, ophyd.Kind.hinted)
-        assert device_signals == signals_from_panel('read_panel')
+        if 'motor_setpoint' in device_signals:
+            # Signal is renamed and not reflected here
+            device_signals.remove('motor_setpoint')
+            device_signals.add('motor')
+        assert device_signals == signals_from_panel('hint_panel')
+
+    def check_read_panel(device):
+        device_signals = signals_from_device(device, ophyd.Kind.normal)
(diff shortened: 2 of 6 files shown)


klauer:
(oops - hit open PR instead of submit image; not quite done yet)

klauer:
If we're all good on how this looks, I'll fix up the tests.

ZLLentz:
I'm not going to give a full review late on a Friday, but I think these look good.

typhos/app.py
```diff
@@ -27,6 +27,8 @@ def launch_suite(suite):
     """Creates a main window and execs the application."""
     window = QMainWindow()
     window.setCentralWidget(suite)
+    window.setWindowTitle(suite.windowTitle())
+    window.setUnifiedTitleAndToolBarOnMac(True)
```
ZLLentz, line 31: I'm surprised to see such a specifically named function in pyqt. Gave me a chuckle.

PR #330, opened 2020-05-20
FIX: Adjust comparison to check for None instead of general check.

hhslepicka:
Take a sit as this one is going to be hard to believe...
@klauer and I chased the damn ghost widgets for a while in the past and it looked like it was because of multiple calls to the creation of the widgets.
After noticing them back when switching templates or refreshing templates via the display settings menu I decided to go after this hunt again.

## Steps to reproduce the issue
- Start a SimDetector
- typhos ophyd.areadetector.SimDetector['{"name":"camera", "prefix":"13SIM1:"}']
- Click at the Display Settings... and select Refresh Templates.
- Notice now the ghost widgets appearing without label or readback.

image

Interestingly, this does not happen with a Fake device:
- typhos --fake ophyd.areadetector.SimDetector['{"name":"camera", "prefix":"13SIM1:"}']

## Description

While testing I was being lead to believe that it was due to the fact that kind was different at component from when the signal was instantiated but this was a red herring.

In fact for my surprise the issue was on the fact that TyphosLabel and TyphosCombobox have different behavior when compared on an if statement.

Here is what I mean:

```ipython

In [1]: from typhos.widgets import TyphosLabel, TyphosComboBox
In [2]: label = TyphosLabel()
In [3]: combo = TyphosComboBox()
In [4]: def test(widget):
   ...:     if widget:
   ...:         return True
   ...:     return False
   ...:

In [5]: test(label)
Out[5]: True

In [6]: test(combo)
Out[6]: False
```

## How Has This Been Tested?
Locally

N/A

## Screenshots
image

### DIFF
--- a/typhos/panel.py
+++ b/typhos/panel.py
@@ -490,7 +490,7 @@ class SignalPanel(QtWidgets.QGridLayout):
             item = self.itemAtPosition(row, col)
             if item:
                 widget = item.widget()
-                if widget:
+                if widget is not None:
                     widget.setVisible(visible)
 
         if not visible or info['signal'] is not None:


hhslepicka:
@klauer any insight into why the comparison would fail for combobox?

klauer:
Style guides seem to say "there's no need to check against `None`, just use `if x:`" - but then you get bitten by these strange issues, making pedantic checks worthwhile.

I'm pretty sure this is due to the iterability (presence of `__len__`) of the combobox versus the label:
```
In [17]: class Test:
    ...:     def __len__(self):
    ...:         return 0
    ...:
    ...:

In [18]: len(Test())
Out[18]: 0

In [19]: bool(Test())
Out[19]: False

In [20]: class Test:
    ...:     def __len__(self):
    ...:         return 1
    ...:

In [21]: len(Test())
Out[21]: 1

In [22]: bool(Test())
Out[22]: True
```

ZLLentz:
I am now very wary of stacking iterability onto larger classes... Data structures should should be small and self-contained. Very interesting, glad to see such a short fix.

Issue #258, opened 2020-03-12
Enhance Disconnected Signal Representation


hhslepicka:
## Expected Behavior
From Tim Maxwell:

> My mistake re: purple of course. Feedback is simply something that doesn't get false hope. Instant visual feedback without mouse over I'd think is preferred. Maybe a broken icon/disconnect of some kind :broken_heart:.

From William Schlotter:
> This animation seems to make me feel that I'm going to see a PV soon.   I'm always disappointed...could it be a red X?

## Current Behavior

Currently the system shows an animation while the signal does not receive the connection callback.

## Possible Solution
As pointed above, one possible solution is to replace the animation with a broken link static representation along with a tooltip to clarify.

## Steps to Reproduce (for bugs)

1. Open a Typhos screen for a disconnected device

## Context

This issue affects users as it gives the hope that the signal will be connected soon and not always that is true.

## Your Environment

Development Environmnet.

ZLLentz:
Perhaps we start with the existing GIF, as there may be cases where we have delayed setup, but then switch to the broken icon after a short timeout?

PR #331, opened 2020-05-20
ENH: Drop loading gif after a certain timeout.

hhslepicka:
## Description
Add a `LOADING_TIMEOUT_MS` and `enable_timeout` options at `TyphosLoading` widget.

## Motivation and Context
Closes https://github.com/pcdshub/typhos/issues/258

## How Has This Been Tested?
Locally

Docstrings

timeout

### DIFF
--- a/typhos/utils.py
+++ b/typhos/utils.py
@@ -166,24 +166,25 @@ def random_color():
 
 
 class TyphosLoading(QtWidgets.QLabel):
+    """
+    A QLabel with an animation for loading status.
+
+    Attributes
+    ----------
+    LOADING_TIMEOUT_MS : int
+        The timeout value in milliseconds for when to stop the animation
+        and replace it with a default timeout message.
+
+    Parameters
+    ----------
+    enable_timeout : bool
+        Whether or not to stop the animation after a timeout value.
+    """
     LOADING_TIMEOUT_MS = 10000
     loading_gif = None
-    """Simple widget that displays a loading GIF"""
+
     def __init__(self, enable_timeout=True, *args, **kwargs):
-        """
-        A QLabel with an animation for loading status.
 
-        Attributes
-        ----------
-        LOADING_TIMEOUT_MS : int
-            The timeout value in milliseconds for when to stop the animation
-            and replace it with a default timeout message.
-
-        Parameters
-        ----------
-        enable_timeout : bool
-            Whether or not to stop the animation after a timeout value.
-        """
         super().__init__(*args, **kwargs)
         self._icon_size = QSize(32, 32)
         if TyphosLoading.loading_gif is None:
--- a/typhos/utils.py
+++ b/typhos/utils.py
@@ -198,7 +198,6 @@ class TyphosLoading(QtWidgets.QLabel):
         self._animation.stop()
         self.setMovie(None)
         self.setText("Loading Timeout")
-        self.setStyleSheet("color: red;")
         self.setToolTip("Could not complete operation after "
                         f"{self.LOADING_TIMEOUT_MS/1000} seconds.")
 
--- a/typhos/utils.py
+++ b/typhos/utils.py
@@ -175,15 +175,11 @@ class TyphosLoading(QtWidgets.QLabel):
         The timeout value in milliseconds for when to stop the animation
         and replace it with a default timeout message.
 
-    Parameters
-    ----------
-    enable_timeout : bool
-        Whether or not to stop the animation after a timeout value.
     """
     LOADING_TIMEOUT_MS = 10000
     loading_gif = None
 
-    def __init__(self, enable_timeout=True, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


typhos/utils.py
```diff
@@ -166,9 +166,24 @@ def random_color():
...
+        ----------
+        LOADING_TIMEOUT_MS : int
+            The timeout value in milliseconds for when to stop the animation
+            and replace it with a default timeout message.
+
+        Parameters
+        ----------
+        enable_timeout : bool
+            Whether or not to stop the animation after a timeout value.
+        """
```
ZLLentz, line 186: Usually we put these doc strings at the class level rather than attached to `__init__`. This should replace the `"""Simple widget that displays a loading GIF"""` text (which should have been at the very top of the class def)

typhos/utils.py
```diff
@@ -178,6 +193,17 @@ def __init__(self, *args, **kwargs):
...
+            QtCore.QTimer.singleShot(self.LOADING_TIMEOUT_MS,
+                                     self._handle_timeout)
+
+    def _handle_timeout(self):
+        self._animation.stop()
+        self.setMovie(None)
+        self.setText("Loading Timeout")
+        self.setStyleSheet("color: red;")
+        self.setToolTip("Could not complete operation after "
+                        f"{self.LOADING_TIMEOUT_MS/1000} seconds.")
```
ZLLentz, line 206: Are there any problems here if the loading widget has already been removed/replaced? Or does QT clean up nicely for timers referencing methods in deleted widgets?

hhslepicka: On local tests it handled it properly. I started with an offline IOC and before the 10s I enabled the IOC, no issues were noticed.

typhos/utils.py
hhslepicka: Addressed via https://github.com/pcdshub/typhos/pull/331/commits/68bb35867b474dd38f19bd8076de99987bffbaa9

hhslepicka:
The linter failure seems weird. I think that changes added into `pcds-ci-helpers` are now triggering a full test of the suite not only the linter. Tests are passing and local check with flake8 is also good. I will investigate the linter job with Tyler.

Should I push fixes to that into this PR or would it be better to have a separated one to deal with that?

ZLLentz:
Also, I'm always 👎 on merging unrelated fixes into the same PR. Life is simpler when you can review one thing at a time (though I know this isn't always feasible for various reasons).

typhos/utils.py
klauer: Seems like it's OK with a bound method:

```python
import qtpy
import qtpy.QtWidgets
import qtpy.QtCore


class Test(qtpy.QtWidgets.QWidget):
    def __init__(self):
        super().__init__()
        qtpy.QtCore.QTimer.singleShot(100, self.test)

        def test2():
            print('inner function')

        qtpy.QtCore.QTimer.singleShot(100, test2)

    def test(self):
        print('test called / bound method')


app = qtpy.QtWidgets.QApplication([])

test = Test()
test.deleteLater()
del test
test = None

app.exec_()
```

Results in: 
```
inner function
```

typhos/utils.py
```diff
@@ -178,6 +193,17 @@ def __init__(self, *args, **kwargs):
...
         self._animation.start()
+        if enable_timeout:
+            QtCore.QTimer.singleShot(self.LOADING_TIMEOUT_MS,
+                                     self._handle_timeout)
+
+    def _handle_timeout(self):
+        self._animation.stop()
+        self.setMovie(None)
+        self.setText("Loading Timeout")
+        self.setStyleSheet("color: red;")
```
klauer, line 205: Should be in the stylesheet file

typhos/utils.py
```diff
@@ -166,9 +166,24 @@ def random_color():
...
+
+        Attributes
+        ----------
+        LOADING_TIMEOUT_MS : int
+            The timeout value in milliseconds for when to stop the animation
+            and replace it with a default timeout message.
+
+        Parameters
+        ----------
+        enable_timeout : bool
```
klauer: Maybe unnecessary; could leave it based on `if LOADING_TIMEOUT_MS > 0`

hhslepicka: Addressed via https://github.com/pcdshub/typhos/pull/331/commits/68ae4b6e9b26f8b11ad40f47aca6c8953d82bcd3

typhos/utils.py
hhslepicka: Addressed via https://github.com/pcdshub/typhos/pull/331/commits/2e344b8798d9fc7cdd34c8190688a1271922fb7e

typhos/utils.py
ZLLentz:
```
test.deleteLater()
del test
test = None
```
It seems like you do not trust the Python garbage collector (probably warranted)

PR #352, opened 2020-06-12
MNT: positioner widget set timeouts

ZLLentz:
## Description

Use current position, velocity, and acceleration signals to pick a motion timeout value of an appropriate size.

## Motivation and Context

Slow moving motors hit a default timeout of 10s at some point in the call stack. This raises the `WaitTimeout` error in the screen, confusing users. This is a straightforward way to define a timeout that is "high enough" to not trip unless there is a problem.

## How Has This Been Tested?

I haven't tested this yet

Docstrings only

### DIFF
--- a/typhos/positioner.py
+++ b/typhos/positioner.py
@@ -171,7 +171,7 @@ class TyphosPositionerWidget(utils.TyphosBase, widgets.TyphosDesignerMixin):
         thread.status_finished.connect(self._status_finished)
         thread.start()
 
-    def _get_timeout(self, value, settle_time):
+    def _get_timeout(self, set_position, settle_time):
         """Use positioner's configuration to select a timeout."""
         pos_sig = getattr(self.device, self._readback_attr, None)
         vel_sig = getattr(self.device, self._velocity_attr, None)
@@ -179,7 +179,7 @@ class TyphosPositionerWidget(utils.TyphosBase, widgets.TyphosDesignerMixin):
         # Not enough info == no timeout
         if pos_sig is None or vel_sig is None:
             return math.inf
-        delta = abs(pos_sig.get() - value)
+        delta = pos_sig.get() - set_position
         speed = vel_sig.get()
         # Bad speed == no timeout
         if speed == 0:
@@ -190,16 +190,17 @@ class TyphosPositionerWidget(utils.TyphosBase, widgets.TyphosDesignerMixin):
         else:
             acc_time = acc_sig.get()
         # This time is always greater than the kinematic calc
-        return delta/speed + 2 * acc_time + abs(settle_time)
+        return abs(delta/speed) + 2 * abs(acc_time) + abs(settle_time)
 
     def _set(self, value):
         """Inner `set` routine - call device.set() and monitor the status."""
         self._clear_status_thread()
         self._last_move = None
+        set_position = float(value)
 
         logger.debug("Setting device %r to %r", self.device, value)
-        timeout = self._get_timeout(value, 5)
-        status = self.device.set(float(value), timeout=timeout)
+        timeout = self._get_timeout(set_position, 5)
+        status = self.device.set(set_position, timeout=timeout)
         self._start_status_thread(status)
 
     @QtCore.Slot()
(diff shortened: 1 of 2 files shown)


typhos/positioner.py
```diff
@@ -385,6 +410,24 @@ def high_limit_travel_attribute(self):
...
+
+    @QtCore.Property(str, designable=True)
+    def acceleration_attribute(self):
+        """The attribute name for the acceleration time signal."""
+        return self._acceleration_attr
+
+    @acceleration_attribute.setter
+    def acceleration_attribute(self, value):
+        self._acceleration_attr = value
+
```
hhslepicka, line 430: Since you are here 😄 ... Should also the limits (switch and travel) be configurable as you did for the velocity and acceleration?

typhos/positioner.py
```diff
@@ -168,13 +171,35 @@ def _start_status_thread(self, status):
         thread.status_finished.connect(self._status_finished)
         thread.start()
 
+    def _get_timeout(self, value, settle_time):
+        """Use positioner's configuration to select a timeout."""
+        pos_sig = getattr(self.device, self._readback_attr, None)
+        vel_sig = getattr(self.device, self._velocity_attr, None)
+        acc_sig = getattr(self.device, self._acceleration_attr, None)
```
hhslepicka, line 178: Are those attributes using cached values at the signal or always doing a get?
Could this cause a problem since this is being calculated at every movement request?

typhos/positioner.py
ZLLentz: Those are already configurable, this snippit is actually just copy/pasted from those and modified.

typhos/positioner.py
ZLLentz: Looking at ophyd, these gets will not used any cached values. Perhaps there is a better way to get these since the screen should already be monitoring them in some way for the display.

typhos/positioner.py
hhslepicka: Ooops... missed that...

typhos/positioner.py
hhslepicka: I think that they will be fine. Sorry my brain is damaged today. If it becomes a big deal we can always refactor later on.

ZLLentz:
Darn, my "pray that the tests just still pass" strategy was marginal at best. I'll see if I can track it down locally.

ZLLentz:
The test freeze is because an exception dialog popup happens during the test, and it is waiting for user input before proceeding.

ZLLentz:
This should pass now. I will take it for a test drive on Monday with a "real" positioner as a final check to make sure I didn't overlook something.

ZLLentz:
The remaining error is a requirements comparison linter. This PR did not update the requirements files. Is this an artifact of our new shared travis config?

klauer:
It's new - https://github.com/pcdshub/pcds-ci-helpers/pull/36#issuecomment-640764060

ZLLentz:
This was not ready yet :( was doing the test I said I wanted to do
```
[2020-06-16 09:12:22] - DEBUG - Thread (139782279919424 - MainThread ) - typhos.positioner -> Setting device BeckhoffAxis(PLC:TST:MOT:SIM:01, name=BeckhoffAxis) to '300' with timeout 75.00133333333333
[2020-06-16 09:12:22] - DEBUG - Thread (139782279919424 - MainThread ) - typhos.positioner -> Begin showing move in TyphosPositionerWidget
[2020-06-16 09:12:32] - ERROR - Thread (139778402006784 - Dummy-23 ) - typhos.status -> WaitTimeoutError: Status MoveStatus(done=False, pos=BeckhoffAxis, elapsed=10.1, success=False, settle_time=0.0) did not complete in 10.0 seconds
[2020-06-16 09:12:32] - DEBUG - Thread (139778402006784 - Dummy-23 ) - typhos.status -> Emitting finished: WaitTimeoutError('Status has not completed yet.',)
[2020-06-16 09:12:32] - DEBUG - Thread (139782279919424 - MainThread ) - typhos.positioner -> Completed move in TyphosPositionerWidget (result=WaitTimeoutError('Status has not completed yet.',))

```

klauer:
Saw tests pass, sorry. Will need follow-up PRs for that + requirements.

Requirements testing overall is busted, though, so we'll probably be ignoring it for a while.

PR #357, opened 2020-06-23
Variety metadata + test IOC

klauer:
## Description

Variety metadata adds the possibility for distinguishing different types of signals from one another.
These different varieties of signals map differently onto typhos widgets.

This PR adds the framework and outlines examples for handling all of the currently defined varieties (pairing with https://github.com/pcdshub/pcdsdevices/pull/463)

While this adds a considerable amount of magic by way of decorators, I believe the sheer number of metadata keys, varieties, and classes justifies this. These additional decorators also allow for automatic annotation of class docstrings for documentation generation and such.

Note: For signals without variety designation (99.995% of them currently), this PR should have no effect.

## How Has This Been Tested?
Almost entirely locally. This could be considered an experimental feature.

Local testing has been with `variety_ioc` included in the test suite. It can be launched by way of: `TYPHOS_DEBUG=1 typhos 'variety.MyDevice[{"prefix":"variety:"}]'`

Widget docstrings + some sphinx docs updates.

## Work-in-progress status
- [x] Link all varieties to widgets (placeholders for new ones)
- [ ] Implement new widgets (may be follow-up PRs for even basic functionality)
   - [x] Multiline text edit
   - [x] Support bitmask setpoints (not ideal, but it works)
   - [x] Support general tweakable (non positioner; basic work is in)
   - [ ] Histogram?
- [ ] Respect variety metadata as noted in planning document (will make an attempt here, may be follow-up PRs)
   - [ ] New widget functionality to support variety metadata
   - [x] Tweakable/scalar-range: delta source from signal/value
   - [x] Password prompts, confirmation for pushbuttons
   - [ ] Password prompts, confirmation for other widgets? Not natively supported by PyDM
- [x] Mixin / hook for generically setting metadata for new widgets
- [x] Use provided caproto IOC for interactive testing
- [ ] Use provided caproto IOC for continuous integration (may be in a follow-up PR)

## Screenshots
Updating a preview as I go through this (variety IOC; `typhos 'variety.MyDevice[{"prefix":"variety:"}]'`):

<details>
<summary>Attempt 1</summary>
image 
</details>

<details>
<summary>Attempt 2</summary>
image
</details>

<details>
<summary>Attempt 2.5?</summary>
image
</details>

<details>
<summary>Attempt 3</summary>
image
</details>

Current attempt:
image

### DIFF
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -2,7 +2,7 @@
 
 import enum
 import logging
-import os.path
+import os
 import pathlib
 
 from qtpy import QtCore, QtGui, QtWidgets
@@ -323,6 +323,11 @@ class TyphosDisplayConfigButton(TyphosToolButton):
 
         base_menu.addSeparator()
         self.create_hide_empty_menu(panels, base_menu)
+
+        if utils.DEBUG_MODE:
+            base_menu.addSection('Debug')
+            action = base_menu.addAction('&Copy to clipboard')
+            action.triggered.connect(display.copy_to_clipboard)
 
         return base_menu
 
@@ -1135,6 +1140,26 @@ class TyphosDeviceDisplay(utils.TyphosBase, widgets.TyphosDesignerMixin,
                 device_cls, 'detailed', utils.DISPLAY_PATHS)
             if not utils.is_standard_template(template)
         ]
+
+    def to_image(self):
+        """
+        Return the entire display as a QtGui.QImage.
+
+        Returns
+        -------
+        QtGui.QImage
+            The display, as an image.
+        """
+        if self._display_widget is not None:
+            return utils.widget_to_image(self._display_widget)
+
+    @Slot()
+    def copy_to_clipboard(self):
+        """Copy the display image to the clipboard."""
+        image = self.to_image()
+        if image is not None:
+            clipboard = QtGui.QGuiApplication.clipboard()
+            clipboard.setImage(image)
 
     @Slot(object)
     def _tx(self, value):
--- a/typhos/tweakable.py
+++ b/typhos/tweakable.py
@@ -28,6 +28,9 @@ class TyphosTweakable(utils.TyphosBase):
 
     init_channel : str, optional
         The channel to be used by the widget.
+
+    Notes
+    -----
     """
 
     ui_template = utils.ui_dir / 'tweakable.ui'
--- a/typhos/tweakable.py
+++ b/typhos/tweakable.py
@@ -1,0 +1,85 @@
+"""
+Tweakable value widget.
+
+Variety support pending:
+- everything
+"""
+import logging
+
+import qtpy
+from qtpy import QtCore
+
+from . import utils, variety
+
+logger = logging.getLogger(__name__)
+
+
+@variety.uses_key_handlers
+@variety.use_for_variety_write('scalar-tweakable')
+class TyphosTweakable(utils.TyphosBase):
+    #  TODO rearrange package: widgets.TyphosDesignerMixin):
+    """
+    Widget for a tweakable scalar.
+
+    Parameters
+    ----------
+    parent : QWidget
+        The parent widget.
+
+    init_channel : str, optional
+        The channel to be used by the widget.
+    """
+
+    ui_template = utils.ui_dir / 'tweakable.ui'
+    _readback_attr = 'readback'
+    _setpoint_attr = 'setpoint'
+
+    def __init__(self, parent=None, init_channel=None, variety_metadata=None,
+                 ophyd_signal=None):
+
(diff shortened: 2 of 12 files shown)


klauer:
OK, so: there is a new lead for UI/UX (yay, @ZLLentz!)

I don't want to bring this PR too far forward without direction, and moving this PR much further would easily be in 'stepping on toes' territory.

For signals _without_ variety designation (99.995% of them currently), this PR _should_ have no effect.

With the above in mind, and noting that this PR is incomplete, I'll fix up the tests and mark this as ready for review.

typhos/display.py
```diff
@@ -1136,6 +1141,26 @@ def _get_specific_screens(cls, device_cls):
...
+        if self._display_widget is not None:
+            return utils.widget_to_image(self._display_widget)
+
+    @Slot()
+    def copy_to_clipboard(self):
+        """Copy the display image to the clipboard."""
+        image = self.to_image()
+        if image is not None:
+            clipboard = QtGui.QGuiApplication.clipboard()
+            clipboard.setImage(image)
```
ZLLentz, line 1162: This isn't the point of the PR but it looks really useful

typhos/tweakable.py
```diff
@@ -0,0 +1,88 @@
...
+
+from . import utils, variety
+
+logger = logging.getLogger(__name__)
+
+
+@variety.uses_key_handlers
+@variety.use_for_variety_write('scalar-tweakable')
+class TyphosTweakable(utils.TyphosBase):
+    #  TODO rearrange package: widgets.TyphosDesignerMixin):
```
ZLLentz, line 20: We should transition any TODOs over to GH issues

typhos/utils.py
```diff
@@ -1048,3 +1050,65 @@ def get_text(item):
...
+    component : ophyd.Component
+        The component, if available.
+    """
+    if obj.parent is None:
+        return None
+
+    return getattr(type(obj.parent), obj.attr_name, None)
+
+
+def get_variety_metadata(cpt):
```
ZLLentz, line 1075: I see there are some duplicated functions here. I assume this is to avoid weird dependency chains?

typhos/display.py
klauer: :+1: Example is at the bottom of the PR description. Once that display got to be beyond 1 screen high, the tool became a necessity.

typhos/tweakable.py
klauer: OK, if you agree with them as-is, I'll open up issues with further details.

typhos/utils.py
klauer: Yes - we don't want typhos to depend on pcdsdevices. The only common place, otherwise, would be to move variety metadata stuff to either pcdsutils (or less likely, ophyd itself).

typhos/widgets.py
```diff
@@ -256,16 +277,34 @@ def unit_changed(self, new_unit):
...
+        super().unit_changed(new_unit)
+        default = (self.displayFormat == DisplayFormat.Default)
+        if new_unit.lower() in EXPONENTIAL_UNITS and default:
+            self.displayFormat = DisplayFormat.Exponential
+
+
+@use_for_variety_read('array-nd')
+@use_for_variety_read('command')
+@use_for_variety_read('command-enum')
+@use_for_variety_read('command-proc')
```
ZLLentz, line 292: I think the decorator stack is a pretty good solution to categorizing/registering these widgets

klauer: Thanks! (and phew, I was a bit worried I got _too_ carried away here with magic...)

typhos/tweakable.py
```diff
@@ -0,0 +1,88 @@
+"""
+Tweakable value widget.
+
+Variety support pending:
+- everything
```
ZLLentz, line 5: Is there some complexity threshold by which a widget gets its own file vs being in `typhos.widgets`?

typhos/widgets.py
ZLLentz: I think we have worse offenders lying around when it comes to magic... Magic is only a real problem when it makes the code hard to read. I think decorators like these make it easier to read.

typhos/tweakable.py
klauer: Not _really_. I prematurely started moving these out to separate files, expecting that with more metadata support they would balloon much larger than they currently are. 

It'd probably be advisable to move widget source around in the future (a topic for another issue, I suppose).

typhos/tweakable.py
klauer: Issues created. TODOs remain in place, but can be removed as the issues are resolved.

klauer:
OK, here goes nothing...

PR #370, opened 2020-07-22
MNT: pass down kwarg correctly

ZLLentz:
## Description

The kwarg in this method chain is dropped. This PR picks it back up.

## Motivation and Context

This code path is used in lightpath.

## How Has This Been Tested?

Lightpath loads more things now.


klauer:
Down the line, should look into using the connection callback utilities in lightpath as we do in typhos...

Issue #359, opened 2020-07-02
Use some Typhos widgets in templates


klauer:
1. Prefer `TyphosLineEdit` over `PyDMLineEdit` in templates (`TyphosLineEdit` retains setpoint history)
    * `positioner.ui` uses `PyDMLineEdit` for setpoint, tweak

Issue #342, opened 2020-05-26
Create new screens repository (repositories?)


klauer:
Option 1

* ophyd built-in devices remain in repo `typhos`
* Repo `pcds-typhos-screens` (pcds-specific screens)

Option 2

* Repo `typhos-screens` (ophyd built-in devices, useful for other facilities)
* Repo `pcds-typhos-screens` (pcds-specific screens)

Option 3
* ?

Somewhat related to https://github.com/pcdshub/typhos/issues/336

klauer:
Also somewhat related - a 'screen' directory entrypoint for pydm (typhos) might be useful to reduce the amount of environment variable directories that have to be set. (Similar to https://github.com/pcdshub/pyqt-designer-plugin-entry-points/ if we ever get it off the ground)

klauer:
Might also consider submodules (as awful as they are) for screen repositories within `typhos`.

`typhos/ui/{submodules}`
Then, we could even further break things up if desirable: `typhos-area-detector-screens`

ZLLentz:
Seems like you went with option 2, but with a rename: repo `typhos-displays` (and later we'll likely add `pcds-typhos-displays`)

I think it is correct to keep them separated like this.

klauer:
No real decision was made - I just needed a place to toss the areadetector screens in case someone wanted to try them.

klauer:
We have since made:

https://github.com/pcdshub/vacuumscreens
and
https://github.com/pcdshub/ecs-pydm-ui
and we have displays inside pcdsdevices:
https://github.com/pcdshub/pcdsdevices/tree/master/pcdsdevices/ui

Issue #311, opened 2020-05-15
User-preferred templates


klauer:
- [ ] Prefer template menu option per screen type (embedded, detailed, engineering)
- [ ] Persist preferred templates
- [ ] User preference priority just below macro specification

Issue #339, opened 2020-05-22
PositionerWidget fails for StatePositioner


ZLLentz:
## Expected Behavior

- Unsure, might actually need changes in `pcdsdevices`
- This device will want a combo box (it probably doesn't even want the `PositionerWidget`)
- This device will probably be forced into a `PositionerWidget`, being a subclass of `PositionerBase`

## Current Behavior

![image](https://user-images.githubusercontent.com/10647860/82699851-4776a380-9c22-11ea-9cf7-090e283f775d.png)

ZLLentz:
If there's no obvious way to handle a non-float Positioner, I will slot this into the states refactor in pcdsdevices

klauer:
We could refactor the template to check `setpoint.describe()` and hide/show an alternate frame, maybe?

This could be done much more easily if https://github.com/pcdshub/typhos/issues/312 were in, as we could generically say "show me the readback widget here" and "show me the setpoint widget here".

klauer:
Or, failing the above, just using a separate template entirely for the state positioners, despite the code duplication that would be involved in that.

ZLLentz:
Something to note here is that the `StatePositioner` doesn't actually have most of the signals requested by this widget (it just has a `state` signal that is both the read and the write for the position). So the separate template would have almost no similarities with the original.

I was also considering whether or not these devices really classify as positioners (e.g. should they stop depending on `PositionerBase`?) But after re-reading the code in ophyd I think it is appropriate. 

A related thought is that this widget seems to assume that it will find signals similar in function to `EpicsMotor`. Does this mean that the positioner widget is truly an epics motor widget? I think this is somewhat the case, though a few changed attributes make it work for `PVPositioner` too. I don't think there is any way to get it to work for `PsuedoPositioner`.

A lot of these base class definitions in ophyd are based on the functional API rather than on the composite signal names, making it hard to generalize for subclasses in the screens here.

klauer:
I think you'd expect a setpoint combobox, a readback label, and maybe a stop button? So long as the positioner widget hides the unavailable components, it seems like it'd still be OK. (Ideal - no, but sufficient - probably).

`PseudoPositioner` itself wouldn't work, but the detailed view of `PseudoSingle` and `EpicsMotor` individual components _should_ in theory.

Very true regarding functionality over component names.

We may get to a point where we're inserting "component interface" mixins for typhos to pick the right screen (i.e., an interface class that guarantees the existence of certain component names)

[...]

PR #371, opened 2020-07-23
REF: Split templates between core and devices

### LINKED ISSUE #336: Typhos-provided screens not considered "device-specific" screens
* `PositionerBase.*.ui` is provided with typhos, but will not be recommended over `detailed_tree.ui`
* The check is something like: "do we have specific (* not typhos-provided) screens for this class? Is the device complicated enough to suggest a tree?"

Possible solution
------------------
Provide two `.ui` file areas: 
1. `typhos.ui.core` for core templates that have nothing to do with classes
2. `typhos.ui.device` for ophyd-wide device classes (such as `PositionerBase`)
The former would be considered "typhos provided" in the current methodology and the latter would be considered "device-specific"

hhslepicka:
## Description
This PR separates the ui templates into `core`, `devices` and `widgets`.

- `core` should contain the templates that are internal to Typhos for autogenerated screens (detailed_tree, engineering, etc).
- `devices` should contain at least for now templates tailored for specific devices class.
- `widgets` should contain UI files related to Widgets provided by Typhos such as `tweakable`, `positioner`, etc

In order to keep the changes relatively small more work will come at other PRs to address the following somewhat related issues:
- https://github.com/pcdshub/typhos/issues/311
- https://github.com/pcdshub/typhos/issues/359
- https://github.com/pcdshub/typhos/issues/342

## Motivation and Context
Closes #336 

## How Has This Been Tested?
All tests are passing (at least locally).

N/A

Screen generated by: `typhos --fake-device pcdsdevices.pim.PPM['{"prefix":"fooo","name":"PPM_Test"}']`

- with master:
Screen Shot 2020-07-23 at 9 46 25 AM

- with this PR:
Screen Shot 2020-07-23 at 9 45 20 AM


ZLLentz:
I guess there is another problem here in your screenshot where the positioner screen breaks for state positioners (see #339)

hhslepicka:
Working with this branch a bit more after talking with Zach and trying to get a closure also for #339 here.
With the latest changes of this PR we can now see the Positioner and StatePositioner both using the TyphosPositionerWidget to render the screen.
Screen Shot 2020-07-28 at 5 08 50 PM

I still need to test this solution more with a real device but adding this up here to gather input on whether this is a good idea or not.

TODO
------
- [ ] Test with a real device moving it around
- [ ] Add also the StatePositioner.embedded template.

klauer:
Seems reasonable 👍

hhslepicka:
I just tested the new positioner with `im1k1` and issues were fixed.
image

hhslepicka:
This PR is ready for a new round of reviews. @klauer and @ZLLentz 
The only 🐘 in the room is the `Stop` button which we have no clear way to find out programmatically if we need or not to remove it since the method `stop` is implemented at the `PositionerBase`.

klauer:
If it works for both positioner types, I'd say merge at will. 👍

ZLLentz:
Later we will rework `StatePositioner` to not have a `stop` method at all (since there is no real reason it needs to subclass `Positioner`, aside from some refactoring)

Issue #354, opened 2020-06-17
SystemError with TyphosPositionerWidget and Python > 3.7.3


hhslepicka:
## Expected Behavior
No error when using the widget with Python > 3.7.3

## Current Behavior
```python
$ ipython --gui qt
Python 3.7.6 | packaged by conda-forge | (default, Mar 23 2020, 22:45:16)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.14.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from typhos.positioner import TyphosPositionerWidget
Warning: QApplication was created before pyqtgraph was imported; there may be problems (to avoid bugs, call QApplication.setGraphicsSystem("raster") before the QApplication is created).

In [2]: w = TyphosPositionerWidget()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
AttributeError: __getattribute__

The above exception was the direct cause of the following exception:

SystemError                               Traceback (most recent call last)
<ipython-input-2-68004ea9ab29> in <module>
----> 1 w = TyphosPositionerWidget()

~/sandbox/git-pcdshub/typhon-git/typhos/positioner.py in __init__(self, parent)
    148         super().__init__(parent=parent)
    149
--> 150         self.ui = uic.loadUi(self.ui_template, self)
    151         self.ui.set_value.returnPressed.connect(self.set)
    152         self.ui.tweak_positive.clicked.connect(self.positive_tweak)

~/mc/envs/py37/lib/python3.7/site-packages/PyQt5/uic/__init__.py in loadUi(uifile, baseinstance, package, resource_suffix)
    225     from .Loader.loader import DynamicUILoader
    226
--> 227     return DynamicUILoader(package).loadUi(uifile, baseinstance, resource_suffix)
    228
    229

~/mc/envs/py37/lib/python3.7/site-packages/PyQt5/uic/Loader/loader.py in loadUi(self, filename, toplevelInst, resource_suffix)
     70             basedir = os.path.dirname(filename)
     71
---> 72         return self.parse(filename, resource_suffix, basedir)

~/mc/envs/py37/lib/python3.7/site-packages/PyQt5/uic/uiparser.py in parse(self, filename, resource_suffix, base_dir)
   1028             elem = document.find(tagname)
   1029             if elem is not None:
-> 1030                 actor(elem)
   1031         self.finalize()
   1032         w = self.toplevelWidget

[...]

PR #377, opened 2020-07-30
HACK/FIX: patch connectSlotsByName to not raise SystemError

klauer:
Addresses #354 but doesn't _really_ fix it, as it's not under our control.

Screens at least load, but there may be loss of functionality if there are slots marked in the .ui files.
As-is, the positioner works for me on pyqt  - without it, the screens just won't load.

There's a lot of discussion of this issue, in pyqt forums, pydm issues, and so on. Seems like the appropriate fix may be to upgrade PyQt5 to a version that's not yet on conda-forge (5.13.1), downgrade Python, or ... Not sure.

If we can nail down the specifics as to when this happens, this patch could be done on a version-by-version basis. However, it's a minimal hack that shouldn't really affect much, unless we're really looking for that `SystemError` in the first place.

Links to some relevant discussions:
https://github.com/slaclab/pydm/issues/533
http://python.6.x6.nabble.com/New-exception-in-Python-3-7-4-when-bad-code-tries-to-PyQt5-uic-loadUi-td5257823.html
https://github.com/GNS3/gns3-gui/issues/2819


typhos/cli.py
```diff
@@ -184,6 +184,8 @@ def create_devices(device_names, cfg=None, fake_devices=False):
 
 def typhos_run(device_names, cfg=None, fake_devices=False):
     """Run the central typhos part of typhos."""
+    typhos.utils.patch_connect_slots()
```
hhslepicka, line 187: Could this call be moved into the Typhos import?
I ask this because the Qt Designer plugin is also affected by this issue and adding there would probably be a centralized place to address both cases.

hhslepicka:
I can confirm that moving the `patch` function call to `__init__.py` solves both cases.

klauer:
Patch on import is generally bad practice, but I don't see a better way...

hhslepicka:
Maybe in this case we could keep it where it is and also add one at the designer plugin file?
It would probably be less of a bad practice.

klauer:
So where we'd want it to patch:
* Typhos CLI
* Lucid embedded typhos screens
* Designer

Unless you can solve the SystemError, looks like on-import is the best we can do. Let's get @ZLLentz on board before proceeding though.

hhslepicka:
I am trying to address again the SystemError and investing some more time into it.
You are correct... unfortunately the import seems to be the `catch-all` place.

ZLLentz:
I'm on board for this dreadful hack.

klauer:
Tests failing miserably? 😕

Issue #383, opened 2020-08-03
StatePositioner put gets called on startup


ZLLentz:
## Expected Behavior

No signals should be put to when loading a screen.

## Current Behavior

State positioners are put to during startup without the user's input. This raises a text box exception on psbuild (no write access) and can be potentially disruptive if done on a control machine.

## Possible Solution

I couldn't figure out how to resolve this, but I'm submitting a hacky workaround in a minute... I tried changing the sequencing and only connecting to set after we'd already initialized the combobox, but add_device still gets called twice so the set still happens.

## Steps to Reproduce (for bugs)

1. Go to psbuild-rhel7
2. Activate dev conda
3. typhos im1k1

## Context

Opening a state positioner screen on psbuild-rhel7 raises an exception. Opening my dev attenuator screen raises 19 exceptions.

## Your Environment

Latest typhos and dependencies

klauer:
This technically isn't an issue anymore as we have a hack from #384; should we leave this open to indicate the intention to refactor?

ZLLentz:
Yes. I also want to add here that this hasn't fully solved the issue: it seems to do another put when the IOC disconnects and reconnects...

klauer:
Oh no... That's even worse. I can prioritize looking into a more permanent fix, if you don't have the time.

ZLLentz:
For state positioner, it always puts to the last-chosen set state. This value initializes to `Unknown` on the restart and then we get the "can't go to unknown state" message instead of doing any pv puts. It's still very unsettling, and I can't rule out other edge cases where a move fails, we disconnect, someone clears the error, we reconnect, and then automatically try to go to the errored state again...

As far as bugs go this is reasonably high priority, but I'm not sure how it compares with Silke's run 18 hutch python high priority.

klauer:
Fixed in https://github.com/ZLLentz/typhos/commit/2231fa509fb8ad69d54aa84641378b42d0ad1532 as part of #447

PR #384, opened 2020-08-03
MNT: work around init bug

ZLLentz:
## Description

Skip the first put to the combobox

## Motivation and Context

See #383. We're currently putting to signals on startup which can be very disruptive.

## How Has This Been Tested?

It doesn't raise an exception on startup any more. The state device still moves on the first combobox selection action.


ZLLentz:
The benchmark failed, of all things. I'm merging regardless.

PR #389, opened 2020-08-05
FIX: do not change setpoints when QComboBox is scrolled

### LINKED ISSUE #369: ComboBox Widget should not change value if user scroll on top of it
## Expected Behavior
Combobox widgets should not change value if mouse is over and the scroll is hit.

## Current Behavior
Currently on the LUCID screen if the mouse is over one of the set points, scrolling the mouse changes the value. This is undesirable.

## Possible Solution
Install an eventFilter and disable this behavior.

1. Open a Typhos screen with a combobox widget for input
2. Put the mouse over the widget
3. Roll the scroll and notice the value changing

## Context
Requested by Matt Seaberg over SLAC pcds-ui channel at SLACK.

klauer:
## Description
Closes #369 
Plus some `isort`-related spam

## Motivation and Context
Inadvertently changing setpoints while scrolling through screens is not a good thing.

## How Has This Been Tested?
Locally.

### DIFF
--- a/typhos/widgets.py
+++ b/typhos/widgets.py
@@ -8,20 +8,20 @@ import inspect
 import logging
 
 import numpy as np
-import qtawesome as qta
-from pyqtgraph.parametertree import parameterTypes as ptypes
-from qtpy import QtGui, QtWidgets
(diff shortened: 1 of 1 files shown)


klauer:
Seems like we're all in agreement 👍

PR #390, opened 2020-08-05
ENH: add information to the loading widget indicating timeout details

### LINKED ISSUE #355: Feature request: List the missing PV when the timeout
## Expected Behavior
When mousing over or middle-mouse-clicking the `Loading Timout` text, we can see the PV that this item in the Typhos display is looking to communicate with. 

## Current Behavior
Currently there does not appear to be a way to list PVs Typhon cannot find from the Typhos display itself.

## Possible Solution
See Expected behavior

## Context

Having the ability to determine which PVs are missing at a glance is helpful when debugging incorrect IOC builds or offline IOCs. I think this kind of debugging will become more and more used as we move towards operations again.

klauer:
## Description
Closes #355

## Motivation and Context
It's impossible with the current master to determine which PVs are failing to connect.

## How Has This Been Tested?
Locally.

Adds a tooltip, in case the PV itself is rather long:

Also adds a context menu for easy copy/paste-ability:

Copy either a single line or the entire tooltip to the clipboard.

This is a bit lacking in terms of UX discoverability, but I think it's a reasonable trade-off to weird reflowing issues we'd get listing the PVs in the timeout message. Agree?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,28 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.5.0
+    hooks:
+    -   id: no-commit-to-branch
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-ast
+    -   id: check-case-conflict
+    -   id: check-json
+    -   id: check-merge-conflict
+    -   id: check-symlinks
+    -   id: check-xml
+    -   id: check-yaml
+        exclude: '^(conda-recipe/meta.yaml)$'
+    -   id: debug-statements
+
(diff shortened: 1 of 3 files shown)


klauer:
There's the following in PyDM:
https://github.com/slaclab/pydm/blob/d3295e823bd56c8774be970facaa21234029cc20/pydm/widgets/base.py#L650-L687

Middle click on the widget, and it'll show you the address info tooltip _and_ copy it to the clipboard.
It suffers from (arguably) worse discoverability than the above, copying the old EDM functionality, and penalizes those working on certain laptops with only left/right mouse buttons.

I could see what adding a context menu for this functionality looks like if you want. I'm 50/50 on whether it's a good thing to do, despite my obvious dislike of the middle-click behavior.

ZLLentz:
Let's think on it some more. I'm also not sure if it's what we should do, though I do believe that a modern ui should never assume the user has a middle click.

klauer:
Sounds good. Will move forward with this one for now.

PR #392, opened 2020-08-05
MNT: show byte indicator as readback for "command"; default button text

### LINKED ISSUE #382: Use bitmask indicator for readback of command variety
Or perhaps no widget at all?

These are not particularly nice (variety=command):

image

klauer:
## Description
Closes #382

Command buttons and their readback are poorly represented in auto-generated typhos screens.
This takes one small step in the direction of rectifying that - see the screenshots for details.

Before:
image

After:
image

### DIFF
--- a/typhos/widgets.py
+++ b/typhos/widgets.py
@@ -14,7 +14,6 @@ import pydm.widgets.base
 import pydm.widgets.byte
 import pydm.widgets.enum_button
 import qtawesome as qta
-
 from ophyd.signal import EpicsSignalBase
 from pydm.widgets.display_format import DisplayFormat
 from pyqtgraph.parametertree import parameterTypes as ptypes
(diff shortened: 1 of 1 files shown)


ZLLentz:
Also: how do those byte indicators behave? Alarm color if true, grey if false?

klauer:
It looks like custom button text may have been an oversight in the variety schema. But we should be able to get some text from the channel enum, I think. That needs more investigation.

So:
When used as a byte indicator, the colors are configurable with variety metadata (via 'style' / 'on_color').
When used here, it actually exposes a bit of a flaw in the system: 
The readback and setpoint - despite sharing the same "variety" - have entirely different widgets and require different metadata for customization. For example, `on_color` should apply to the indicator, but not the pushbutton.

Not sure of the best path forward yet...

klauer:
Oh, and to more directly answer your question:
https://github.com/slaclab/pydm/blob/d3295e823bd56c8774be970facaa21234029cc20/pydm/widgets/byte.py#L85-L86
On/off are specified by default RGB values in the widget.

ZLLentz:
I think it is OK for the readback and the setpoint widgets to draw from different parts of the metadata to make these sorts of decisions. But this PR is solid as-is, even we raise some follow-up questions.

klauer:
OK, sounds good. Away we go...

PR #398, opened 2020-08-13
TST: try moving tests into the typhos package

### LINKED ISSUE #397: ClobberError conflict with hklpy
When installing [hklpy](https://github.com/bluesky/hklpy) and [typhos](https://github.com/pcdshub/typhos) as part of a [bluesky environment for the APS](https://github.com/BCDA-APS/use_bluesky/blob/master/python_installation/environment_2020_9.yml), two `ClobberError` reports were generated (https://github.com/BCDA-APS/use_bluesky/issues/63) during the environment creation process (on linux x86_64, linux mint).  

```
ClobberError: This transaction has incompatible packages due to a shared path.
  packages: nsls2forge/noarch::hklpy-0.3.14-py_0, conda-forge/noarch::typhos-1.0.2-py_0
  path: 'lib/python3.8/site-packages/tests/__init__.py'


ClobberError: This transaction has incompatible packages due to a shared path.
  packages: nsls2forge/noarch::hklpy-0.3.14-py_0, conda-forge/noarch::typhos-1.0.2-py_0
  path: 'lib/python3.8/site-packages/tests/__pycache__/__init__.cpython-38.pyc'
```

Does this package install into `tests`?  Why?  Same question asked of hklpy: https://github.com/bluesky/hklpy/issues/22

klauer:
Closes #397 

Tests are being installed separately in their own `tests` package.

We should see what other PCDS Python projects still have this pattern, as it's definitely a problem...


ZLLentz:
Tests being installed as a package is really bad... tests being installed as a submodule of typhos is clearly better.
I'm assuming this issue happens because we have a top-level `__init__.py` in the `tests` module? Therefore making it a "module"?

My perspective is that tests shouldn't be installed at all, and then useful items from package tests should make it into the main package.

klauer:
Yes, it's installed due to setuptools finding it as a package.

bluesky, ophyd, caproto all keep their tests importable and can be re-run with pytest even when installed.
That said, I see this as just a difference of opinion and not a technical issue.

If we disagree strongly enough that you absolutely don't want the tests in the package, please take over the PR and refactor it as you please.

ZLLentz:
No, I don't disagree strongly enough to care about whether tests are shipped with a lib or not.
Tests being runnable as-installed is an interesting feature :shrug:

klauer:
OK, sounds good. Yeah, overall importable tests are of questionable utility. I rather like that it keeps code structured in one package, and not just orphaned in the source tree.

This is ready for review then, as we're back to pre-PR test failures.

prjemian:
Thanks!

klauer:
Thanks for reporting it, @prjemian 👍

PR #401, opened 2020-08-19
BLD: include new ui file location in manifest

ZLLentz:
## Description

ui files moved in a recent PR, the manifest needs to be adjusted

## Motivation and Context

the as-built package is not shipped with the ui files

## How Has This Been Tested?

local conda build

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,6 @@
 include versioneer.py
 include typhos/_version.py
-include typhos/ui/*.ui
+include typhos/ui/*/*.ui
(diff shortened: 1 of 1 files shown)


ZLLentz:
The problem with these is that we run the tests using the cloned version, not the installed version (which is generally just used in the travis for setting up the dependencies.) I'm not sure the best way to resolve this long-term.

klauer:
Good point. We should always be testing with installed versions instead of the source tree, especially now that we're including tests in our repositories in general.

PR #412, opened 2020-12-18
FIX: add ophyd.signal.SignalRO to the list of read-only classes

### LINKED ISSUE #411: AT2L0 num_inserted is shown as read/write in typhos
(Not sure if this belongs in here or typhos - just reporting what I see for now)

Looks like typhos misclassifies`InternalSignal` as writable in my local testing:
https://github.com/pcdshub/pcdsdevices/blob/1cedbe3fd7d1a3513cdafbe0bb7d97795023b36c/pcdsdevices/attenuator.py#L736

image

klauer:
## Description
`ophyd.signal.SignalRO` is another read-only signal variant that is not currently checked in typhos.

Rather than adding that class to the list of classes to check in addition to our own `SignalRO`, this PR makes those two classes one in the same - if available in ophyd, of course.

## Motivation and Context
Closes #411 

## How Has This Been Tested?
Test suite which leverages the built-in `SignalRO` continues to work locally.

Before:
image

After:
image

n/a

### DIFF
--- a/typhos/utils.py
+++ b/typhos/utils.py
@@ -60,17 +60,22 @@ def _get_display_paths():
 DISPLAY_PATHS = list(_get_display_paths())
 
 
-class SignalRO(ophyd.sim.SynSignalRO):
-    def __init__(self, value=0, *args, **kwargs):
-        self._value = value
-        super().__init__(*args, **kwargs)
-        self._metadata.update(
-            connected=True,
-            write_access=False,
-        )
+if hasattr(ophyd.signal, 'SignalRO'):
+    SignalRO = ophyd.signal.SignalRO
+else:
+    # SignalRO was re-introduced to ophyd.signal in December 2019 (1f83a055).
+    # If unavailable, fall back to our previous definition:
+    class SignalRO(ophyd.sim.SynSignalRO):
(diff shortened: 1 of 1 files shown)


ZryletTC (approved):
Makes sense and looks good. Can you explain why it's set up as a subclass of SynSignalRO instead of Signal (like ophyd.signal does it) though?

klauer:
@ZryletTC it was switched over at some point upstream during a refactor: https://github.com/bluesky/ophyd/commit/7e8d56b71122f2d8a12399f37824be0326a2e666

I think it's acceptable not to care here as we don't rely on this `SignalRO` in other libraries, and anything that depends on typhos should not either (downstream applications/libraries' dependency for signal-related classes should be `ophyd` itself).

ZLLentz (approved):

ZLLentz:
This build has a reproducible freeze on travis on python 3.8 after the `typhos/tests/test_suite.py::test_hide_embedded_display` test. I'm going to merge anyway for the end of year tag and make a note in the issues.

https://travis-ci.com/github/pcdshub/typhos/jobs/462104493#L3279

PR #425, opened 2021-02-22
CI: Fix docs build

ZLLentz:
## Description

Set ipython version to the highest on conda forge that exists for python 3.6 in hopes of fixing the docs CI
Note: we should consider moving our docs builds to 3.8 or even 3.9

## Motivation and Context

Docs builds started failing because the ipython sphinx extension has a bug in a random old build, and that build gets picked with high priority whenever it is allowed for some reason.

## How Has This Been Tested?

The CI is the test, but I created a bunch of environments offline

This is docs

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,7 +3,7 @@ codecov
 doctr
 doctr-versions-menu
 flake8
-ipython
+ipython>=7.16
 line_profiler
 pytest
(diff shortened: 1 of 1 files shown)


ZryletTC (approved):
I'm all for moving to 3.8 (or 3.9?) but this looks like a fine fix in the meantime 👍

PR #431, opened 2021-03-23
MNT: untested displayformat fix

klauer:


klauer:
Tyler reports this still doesn't work; not ready for review/merging until we look further into it.

PR #436, opened 2021-04-02
MNT: test suite compatibility with pydm v1.11.0

ZLLentz:
## Description

Work-in-progress changes to get the test suite to pass with `pydm=1.11.0`

## Motivation and Context

The test suite freezes up because of an uncaught exception. I'm not really sure what changed, I think it boils down to some assumptions made in the test here that may not even impact daily usage.

## How Has This Been Tested?

Yes

Instant release notes


klauer:
This ends up being a bug I introduced - albeit a long time ago:

Forced templates from the CSS end up just being strings, instead of `pathlib.Path`:
https://github.com/pcdshub/typhos/blob/fb75acc81f86772f74d7b4b724b20dacd133bcf6/typhos/display.py#L926-L935

Which then causes this to have a string:
https://github.com/pcdshub/typhos/blob/fb75acc81f86772f74d7b4b724b20dacd133bcf6/typhos/display.py#L830-L831

And therefore the `load_template` to fail when looking at a string.

ZLLentz:
There are also additional things that break here with latest pydm, this is just the first. Maybe there's only one more, maybe there are many more.

klauer:
I reproduced the test_cli failure _once_ locally on the first try, fixed what appeared to be another str/path issue and pushed 1c61b01.

I neglected to get the exception text, so I went to reproduce on 1c61b01~1 (prev commit), and 100 test runs later, I can't reproduce it.

So, looks like that's not it.

klauer:
Huh, running the _whole_ test suite repeatably causes it. State is leaking out of the others. Now:

With the latest commit:
```
  File "/Users/klauer/Repos/typhos/typhos/display.py", line 839, in load_best_template
    widget = self._load_template(template)
  File "/Users/klauer/Repos/typhos/typhos/display.py", line 912, in _load_template
    return loader(str(filename), macros=self._macros)
  File "/Users/klauer/Repos/pydm/pydm/display.py", line 112, in load_ui_file
    f = macro.substitute_in_file(uifile, macros)
  File "/Users/klauer/Repos/pydm/pydm/utilities/macro.py", line 28, in substitute_in_file
    template = template_for_file(file_path)
  File "/Users/klauer/Repos/pydm/pydm/utilities/macro.py", line 46, in template_for_file
    with open(file_path) as orig_file:
```

klauer:
Looks like the exception notifier mechanism in PyDM turned into a threaded queue-based mechanism in March of last year. 

I don't have an explanation as to why our tests hadn't hit that code in the same way, but I think the solution here is to acknowledge that "yes, test.ui does not exist; please don't raise to operator". That should fix the suite.

... but we'll see. I'll anxiously anticipate a ✅ when I look back at this on Monday....

klauer:
TimeChart and its invalid `fontawesome` alias are causing crashes here, of all things.

ZLLentz:
Ok, just waiting for the conda-forge machinery on timechart now before we can try again

ZLLentz:
Timechart patch is up, so I restarted the build

ZLLentz:
The conda build decided to use the old version again, so I changed the meta.yaml here...

ZLLentz:
Once again, despite definitely being uploaded, neither my local terminal nor travis's can see the new timechart version on conda-forge. Is anaconda.org caching request responses for long periods of time?

ZLLentz:
It seems like it can take up to 1 hour after a successful upload to anaconda.org for a package to be available for download. Re-running.

ZLLentz:
Tests are passing, I'm going to include a release notes page in this PR as well so we can get it all out in one

klauer:
👏 That was a lot more than expected.

docs/source/release_notes.rst
```diff
@@ -2,6 +2,21 @@
  Release History
 =================
 
+v1.1.6 (2020-04-05)
```
klauer, line 5:
```suggestion
v1.1.6 (2021-04-05)
```
😄

klauer (approved):

typhos/tests/test_cli.py
```diff
@@ -38,6 +39,9 @@ def test_cli_no_entry(monkeypatch, qtbot, happi_cfg):
 
 def test_cli_stylesheet(monkeypatch, qapp, qtbot, happi_cfg):
     monkeypatch.setattr(QApplication, 'exec_', lambda x: 1)
+    monkeypatch.setattr(
+        pydm.exception, 'raise_to_operator', lambda *_, **__: None
+    )
```
ZLLentz, line 44: This makes a lot of sense. In a future edit it might make sense to direct raise_to_operator directly to raise inside the test suite.

docs/source/release_notes.rst
ZLLentz: :brain: :potato:

typhos/display.py
```diff
@@ -834,14 +834,24 @@ def load_best_template(self):
...
                 logger.exception("Unable to load file %r", template)
                 # If we have a previously defined template
                 if self._current_template is not None:
                     # Fallback to it so users have a choice
-                    widget = self._load_template(self._current_template)
+                    try:
+                        widget = self._load_template(self._current_template)
+                    except Exception:
+                        logger.exception(
+                            "Failed to fall back to previous template",
```
klauer, line 849: Wait -
```suggestion
                            "Failed to fall back to previous template: %s",
```

🥔 🧠

PR #440, opened 2021-05-11
ENH: add pushbutton for other apps to open Typhos suites

ZLLentz:
## Description

Add `typhos.related_display.TyphosRelatedSuiteButton`, a `QPushButton` that will open up a `TyphosSuite`.

The suite will appear at the location of the button and layer itself on top of the window. If the suite has been moved elsewhere on the screen, the button will move the previous suite back to the button rather than create a new one. When the created suite window is closed, it will be kept and re-used for the next time the button is clicked.

It has the following properties that can be modified in designer:
- `happi_names`, a list of happi device names to include in the suite. This is used instead of the `happi` channel because we need to be able to specify many names rather than just one.
- `happi_cfg`, a string that points to an alternate `happi` config file, if you want to use a different database than is configured by your environment variables.
- `preload`, a boolean that, when set to `True`, will load the suite when the button is loaded rather than when it is clicked.

It can also be used with the `add_device` method like other `typhos` widgets.

## Motivation and Context

This is a generic way to add a expert screen/related display button to a pydm screen that generates and opens up a typhos screen. Some devices have related devices that could be included using this button in a template, some embedded typhos widgets may want a reference to open up the entire screen.

I have plans to use this as part of a sort of "embedded grid" screen style that may end up as part of typhos.

## Known Issues
Some display managers don't respond properly to pyqt's request to un-minimize their windows. In cases where a window is minimized, and the button is pressed, we expect the window to be un-minimized and summoned- so this may cause confusion and may need some more work.

## How Has This Been Tested?

Interactively and unit tests.
image

Docstrings


typhos/related_display.py
```diff
@@ -0,0 +1,121 @@
...
+        if self._suite is None:
+            self.create_suite()
+        parent = self.parent()
+        if parent is None:
+            global_pos = self.pos()
+        else:
+            global_pos = self.parent().mapToGlobal(self.pos())
+        # Different window managers respond differently to the various
+        # methods called here, the chosen sequence was intended for
+        # good behavior on as many systems as possible.
```
klauer, line 80: I think this may be generally useful as a utility method - `typhos.utils.please_focus_on_this_widget_please(self._suite)`

typhos/related_display.py
```diff
@@ -0,0 +1,121 @@
...
+        self._suite.raise_()
+        self._suite.activateWindow()
+        self._suite.setFocus()
+
+    def create_suite(self):
+        """
+        Create and cache the typhos suite.
+        """
+        # Get the devices associated with our designer attributes
+        happi_names = list(map(str, self.happi_names))
```
klauer, line 95: Because it could be a `QStringList`? I'd recommend fixing this up in the property setter rather than on access. That said I'm not 100% sure even that's necessary, as IIRC QString works well enough as a string.

typhos/related_display.py
```diff
@@ -0,0 +1,121 @@
...
+        if happi_names:
+            happi_cfg = str(self.happi_cfg)
+
+            if not happi_cfg:
+                happi_cfg = None
+
+            happi_client = Client.from_config(cfg=happi_cfg)
+            items = []
+            for name in happi_names:
+                search_result = happi_client.search(name=name)[0]
```
klauer, line 105: A bit of error handling may be advisable here, just in case somebody mucks with the database. 

Up to you, but maybe also consider breaking this portion out into `@property def happi_devices(self) -> list[Device]` or something?

typhos/related_display.py
```diff
@@ -0,0 +1,121 @@
...
+
+            with no_device_lazy_load():
+                device_namespace = load_devices(*items, threaded=True)
+
+            devices = [getattr(device_namespace, name) for name in happi_names]
+        else:
+            devices = []
+
+        # Extend with devices from calls to self.add_devices
+        devices.extend(self.devices)
```
klauer, line 116: Is this the sensible ordering, or the opposite way? (asking, not implying either way)

typhos/related_display.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        If True, we'll create the suite ahead of time.
+        """
+        return self._preload
+
+    @preload.setter
+    def preload(self, exec_preload):
+        self._preload = exec_preload
+        if self._preload and not is_qt_designer():
+            self.create_suite()
```
klauer, line 58: `for i in range(10): widget.preload = True`
Not a likely scenario, but probably something to guard against

klauer (approved):
Looking good! Great functionality addition, I think. Just some optional code suggestions for you to weigh up.

typhos/related_display.py
ZLLentz: Good catch, I'll resolve by adding a condition that checks if we've already created the suite.

typhos/related_display.py
ZLLentz: That's fair, I'll come up with a slightly shorter name and pull it into utils

typhos/related_display.py
ZLLentz: Yes, that's the reason. I'll experiment a bit with simplifying this in the property setter and/or removing the handling.

typhos/related_display.py
ZLLentz: I think error handling is needed even if nobody changes the database- I'm not even checking for typos! I'll do some work here.

I think you may be right about breaking it out, but I think it's doing a bit too much lifting to be a property

typhos/related_display.py
ZLLentz: I don't know, I had the same thought last night when I was working on it but decided to just "end and push". I suspect there are a few ways to do this part more cleanly regardless of the ordering.

ZLLentz:
I've incorporated all suggestions into the module, it's a bit cleaner now

klauer:
Looks good to me 👍 Merge at will!

PR #441, opened 2021-06-23
ENH: Typhos Alarm Widgets

ZLLentz:
## Description

Create some small indicator widgets that will represent a summary of the alarm state of an ophyd device.

image

These can be initialized in three ways:
- Using the `.add_device` method.
- By setting the channel to `happi://device_name`
- By setting the channel to `ca://one_pv_name`

If seeded from a device, the `kindLevel` property will be used to determine which signals to consider. We can consider all hinted signals, all normal and hinted signals, all normal, hinted, and config signals, or all signals.

In addition, this PR cleans up the signal pydm backend to use the new metadata callbacks, because I needed signals to have proper alarm states to test this properly and to use it properly with real signals.

To-do:
- [x] Make the widgets work interactively with some test IOC
- [x] Fix broken tests
- [x] Handle non-EPICS signals
- [x] Include these widgets in the designer
- [x] Make sure you can use these in a reasonably useful way from designer (maybe need to add a parameter for extra PVs?)
- [x] Add tests for all of the shapes and for various example devices
- [x] Make the tests pass
- [x] Add mouseover text that shows the alarming/disconnected PVs
- [x] Do some look and feel tests

## Motivation and Context

Lucid's grid could use some alarm states, and it'd be better to define a re-usable widget here.

## How Has This Been Tested?

Unit tests and interactive tests. All features have unit tests except for "add multiple devices to one widget" and "show alarming signals on mouse over".

The tests passed locally.

Only in docstrings so far...


ZLLentz:
This is as far as I'll get before my vacation- feel free to let it sit here if nobody asks for it.

typhos/alarm.py
```diff
@@ -0,0 +1,380 @@
...
+from .plugins import register_signal
+from .utils import (channel_from_signal, get_all_signals_from_device,
+                    TyphosObject)
+from .widgets import HappiChannel
+
+
+logger = logging.getLogger(__name__)
+
+
+class KindLevel:
```
klauer, line 26: Entirely optional, but consider using: https://github.com/zllentz/typhos/blob/d9370c48e0c1c2c6356191fe9f1569c213e2a0d2/typhos/display.py#L24-L32

We had talked about this the other day, and I _knew_ there was something in some package, but I couldn't find it then. Benefit is that you can use `IntEnum` for Python friendliness but also support Q_ENUM without issue.

typhos/alarm.py
```diff
@@ -0,0 +1,380 @@
...
+            self._channels = [channel]
+            # Connect the channel to the HappiPlugin
+            if hasattr(channel, 'connect'):
+                channel.connect()
+
+    def _tx(self, value):
+        """Receive information from happi channel"""
+        self.add_device(value['obj'])
+
+    def init_alarm_state(self):
```
klauer, line 145: So nitpicky I am regretting typing this, but perhaps `reset_alarm_state` instead?

typhos/alarm.py
```diff
@@ -0,0 +1,380 @@
...
+            PyDMChannel(
+                address=addr,
+                connection_slot=partial(self.update_connection, addr=addr),
+                severity_slot=partial(self.update_severity, addr=addr),
+                )
+            for addr in channel_addrs]
+
+        self.device_channels[device.name] = channels
+        for ch in channels:
+            self.addr_channels[ch.address] = ch
```
klauer, line 204: I think there's arguments for/against the separate dictionaries as show here, but just as a refactor consideration:

`addr_info[ch.address] = {"channel": ..., "connected": ..., "severity": ...}` - or using a dataclass to hold that information instead of a dict, even - could be an alternative.

typhos/alarm.py
```diff
@@ -0,0 +1,380 @@
...
+        # super() doesn't work here, some strange pyqt thing
+        default_pydm_event = PyDMPrimitiveWidget.eventFilter(self, obj, event)
+        if default_pydm_event:
+            return True
+        if event.type() == QtCore.QEvent.Enter:
+            alarming = self.show_alarm_tooltip(event)
+            return alarming
+        return False
+
+    def show_alarm_tooltip(self, event):
```
klauer, line 290: This verbose alarm string is so nice!

klauer (approved):
None of my suggestions/thoughts have any implications on the functionality of the code. I think this looks really nice overall, and am excited to try it out.

Considering you're out for the next week, I might just merge this and give a try integrating this with LUCID one day soon... Anything else could be done in a follow-up PR after all.

klauer:
Playing around with this as a user, I have a few ideas about the specifics of the tooltip.

How about showing the ophyd attributes alongside the PV names?
A tooltip for "what device is this" is needed, regardless of connection status.

Perhaps the format could be something like:
```
Device: device_name
----------------------         # <-- show this and below only if an alarm is set
A:B:C_RBV is DISCONNECTED (my_device.c)
```

typhos/alarm.py
ZLLentz: This is worth doing, I'll edit appropriately

typhos/alarm.py
ZLLentz: true

typhos/alarm.py
ZLLentz: Right, that's a good point. I'll think about this and probably refactor a bit

ZLLentz:
I think it's worthwhile to expand the tooltip as you are describing. Another option might be a table?
```
Device: im1l0
--------------------------------------
im1l0.some_pv | SOME:PV | Disconnected
```

klauer:
Table idea is even better! 👍

typhos/alarm.py
ZLLentz: After playing with this a bit, dataclasses are the way to go. I'll probably never use a nested dict ever again.

ZLLentz:
I may abandon the table due to nonsense with how fonts play with the tooltips in qt
This PR is currently broken in designer again and needs some fixing...

ZLLentz:
OK, this is close now, but it needs some CI rehashing because it won't run on python 3.6 now that I added dataclasses in, and all of our CI assumes 3.6

klauer:
`dataclasses` is pip installable for Python 3.6, and a bit smarter on conda: it's a no-op install for Python 3.7+ on conda-forge

ZLLentz:
I'm opting to just update our ci-helpers to 3.8 default (live on the floor)/drop 3.6 (as per NEP 29)/add 3.9 (very hopeful)- does that seem reasonable?

klauer:
I'm totally on board with that 🎆 👍

Issue #114, opened 2018-10-10
Documentation Tool


teddyrendahl:
## Expected Behavior

We should list helpful documentation lists for devices. We can then have an automated tool that either opens a browser or renders the HTML inside the `PyQt` display itself.

klauer:
Think we should move this to lucid, @hhslepicka?

hhslepicka:
This one I think belongs at Typhos since it is related to the devices and not everyone will use Lucid and would benefit from this feature if using Typhos only.

hhslepicka:
@ZLLentz @klauer fishing for some ideas here to implement this in.

Would a new attribute at ophyd Devices so engineers can attribute some sort of documentation link into it be a bad idea?

Do you have other suggestions?

I thought about Happi but that would make it Happi and container specific.

ZLLentz:
Documentation link in `ophyd` is probably :+1:, I think that has broad usage.

I wonder there is a way to use Sphinx to parse a class's docstrings live? Usually in IPython you would use the `?` and `??` operators to view these.

Another idea would be to find some way to get the `url` attribute of the module from `setup.py` and open that webpage.

hhslepicka:
Do we want the class docstrings or more general device user documentation or both?

ZLLentz:
Usually the device user documentation and docstrings are for the python API and IPython usage. Thinking a bit more, what we really want is information about the real device, not the python class... doc link in ophyd could be used to solve that, linking to a confluence page or something.

klauer:
The `numpydoc` module can be used to parse our compliant docstrings. In fact, I had used it in typhos already (`typhos.func`) a while back, so it's already a dependency.

hhslepicka:
@klauer opinions about the user level vs API documentation on what we should display?
I could make both available via a context menu entry:
```
documentation
---------------
Python API
Device Info
```

klauer:
In my opinion, both can be shown at the same time - leave it up to the user to look for what they need.

hhslepicka:
In this case the device documentation link could just simply be included at the docstrings for the ophyd device class.

klauer:
With a webengine widget, you could easily display these sorts of things:

Link to confluence page or direct display thereof
API documentation
<details>
API docs here
</details>

hhslepicka:
@klauer sure! I am trying to find the best place to have the Confluence Page link. 
If it is just a link at the docstrings then we can, as you mentioned before, leave it up tot he user to look for what they need...
[...]

PR #443, opened 2021-06-29
ENH: help widget

klauer:
## Description
A sketch of a help widget, with the start of confluence integration.

Assumes environment variables are set appropriately:
* `TYPHOS_HELP_URL` (confluence, in our case)
* `TYPHOS_HELP_HEADERS` JSON headers for the help get request
* `TYPHOS_HELP_TOKEN` Shorthand for specifying bearer authorization in the get request without the JSON above (we have a general user with read-only access now; access via token not shared password)

Notes:
* If `TYPHOS_HELP_URL` is unset, this feature is a no-op compared to master. You won't see the `?` button even. 
* Confluence tokens are only on dev confluence for now. Not sure when it'll be updated on production.

## Motivation and Context
Initial step toward closing #114 

## How Has This Been Tested?
Interactively.

### New, toggle-able help section
image

### Tooltip and pop-out docstring info
image

image

### Context menu
image

### Confluence integration (click "?" to toggle in-line visibility)
image

### DIFF
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -629,6 +629,7 @@ class TyphosHelpFrame(QtWidgets.QFrame, widgets.TyphosDesignerMixin):
     def show_help(self):
         """Show the help information in a QWebEngineView."""
         if self.help_web_view:
+            self.help_web_view.show()
             return
         self.help_web_view = QWebEngineView()
         if QWebEngineHttpRequest is not None:
@@ -653,21 +654,22 @@ class TyphosHelpFrame(QtWidgets.QFrame, widgets.TyphosDesignerMixin):
         """Hide the help information QWebEngineView."""
         if not self.help_web_view:
             return
-        self.layout().removeWidget(self.help_web_view)
+        self.help_web_view.hide()
         if self._delete_timer is None:
             self._delete_timer = QtCore.QTimer()
             self._delete_timer.setInterval(20000)
             self._delete_timer.setSingleShot(True)
-            self._delete_timer.connect(self._delete_help_if_hidden)
+            self._delete_timer.timeout.connect(self._delete_help_if_hidden)
             self._delete_timer.start()
 
     def _delete_help_if_hidden(self):
         """
-        Slowly react to the help display removal, as setting it back up
-        can be slow and painful.
+        Slowly react to the help display removal, as setting it back up can be
+        slow and painful.
         """
         self._delete_timer = None
         if self.help_web_view and not self.help_web_view.isVisible():
+            self.layout().removeWidget(self.help_web_view)
             self.help_web_view.deleteLater()
             self.help_web_view = None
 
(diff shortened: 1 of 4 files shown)


ZLLentz:
Neat

klauer:
The docs generation stuff is taking shape a bit more, but we're still waiting for the new version of confluence with personal access tokens to be deployed to production before it can be rolled out.

Templates are here:
https://github.com/pcdshub/happi-to-confluence

Page sample screenshots:
image
image

klauer:
This is close enough to go through the review process, I think. Jumping back to happi-to-confluence docs generation now, so no specific rush.

typhos/display.py
```diff
@@ -457,6 +470,228 @@ def mousePressEvent(self, event):
...
+            self._delete_timer = QtCore.QTimer()
+            self._delete_timer.setInterval(20000)
+            self._delete_timer.setSingleShot(True)
+            self._delete_timer.timeout.connect(self._delete_help_if_hidden)
+            self._delete_timer.start()
+
+    def _delete_help_if_hidden(self):
+        """
+        Slowly react to the help display removal, as setting it back up can be
+        slow and painful.
```
ZLLentz, line 668: Does this cause some sort of problem if it isn't deleted? If it has a high startup cost, we could consider just not deleting it until the screen is closed.

ZLLentz (approved):
I think this looks really good- I went through line-by-line but didn't have much to say. What values should we set at LCLS for the environment variables you've introduced here?

typhos/display.py
klauer: Just that it could take up an exorbitant amount of memory (imagine a browser tab for every typhos device in lucid). I thought this was a reasonable "clean up if it seems like the user isn't going to click show-help again in a few seconds".

I didn't investigate too much, though, so it's more based on a hunch.

klauer:
As for the environment variables, testing is now working with the following:
```
# export TYPHOS_HELP_URL="https://confluence-test02.slac.stanford.edu/display/PCDS/{device.name}"
export TYPHOS_HELP_URL="https://confluence-test02.slac.stanford.edu/display/~klauer/{device.name}"
export TYPHOS_HELP_TOKEN=(token you generate from confluence, or the shared ecsapps user one)
```
Without access to the PCDS network, you could just switch the URL to like `https://www.google.com/search?q={device.name}` for fun.

typhos/display.py
ZLLentz: Gotcha, I agree that browser tabs can be memory hogs

typhos/utils.py
```diff
@@ -41,6 +42,16 @@
                                       ('attr', 'component', 'signal'))
 DEBUG_MODE = bool(os.environ.get('TYPHOS_DEBUG', False))
 
+# Help settings:
+# TYPHOS_HELP_URL (str): The help URL format string
+HELP_URL = os.environ.get('TYPHOS_HELP_URL', "").strip()
```
ZLLentz, line 47: Should documentation be added here or elsewhere about what this string looks like and options for it? E.g. like your example it can contain fields like `{device.name}`

klauer: Yes, you're right - let me add some documentation about the documentation (😉) before we move forward with this one.

ZLLentz:
Is this ready to be merged?

klauer:
Yes, if you're happy with 
99d0790

PR #445, opened 2021-07-20
ENH: add Jira bug report dialog [LCLSPC-32]

klauer:
## Description
Prerequisite PR: #443 (code is based on it)

## Motivation and Context
* Add an easy way for users to report issues with devices/user interfaces/etc
* Configured using environment variables, as in the help widget
 
## How Has This Been Tested?
All locally.

See included documentation.

image

image

image

### DIFF
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -484,7 +484,7 @@ class TyphosHelpToggleButton(TyphosToolButton):
         A Qt signal indicating a request to open the Python docstring
         information.
 
-    open_python_docs : QtCore.Signal
+    report_jira_issue : QtCore.Signal
         A Qt signal indicating a request to open the Jira issue reporting
         widget.
 
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -26,6 +26,7 @@ except ImportError:
 from . import cache
 from . import panel as typhos_panel
 from . import utils, widgets
+from .jira import TyphosJiraIssueWidget
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


ZLLentz:
I'd like to review this after #443 is merged for the smaller diff

klauer:
Rebased on master

typhos/display.py
```diff
@@ -483,12 +484,17 @@ class TyphosHelpToggleButton(TyphosToolButton):
         A Qt signal indicating a request to open the Python docstring
         information.
 
+    open_python_docs : QtCore.Signal
```
ZLLentz, line 487:
```suggestion
    report_jira_issue : QtCore.Signal
```

typhos/jira.py
```diff
@@ -0,0 +1,178 @@
...
+        else:
+            self.setWindowTitle("Typhos issue reporting")
+
+    @staticmethod
+    def get_environment():
+        """Get the default environment information."""
+        env = dict(
+            typhos=typhos.__version__,
+            user=getpass.getuser(),
+            hostname=_failsafe_call(socket.getfqdn),
```
ZLLentz, line 83: I think we also want some information on which Python env we're using, I think the most site-agnostic way would be something like:

`python_env=sys.base_prefix`

Which looks like:
```
In [3]: sys.base_prefix
Out[3]: '/reg/g/pcds/pyps/conda/py36/envs/pcds-4.1.6'
```

klauer: Added in 70cd5e8

typhos/display.py
klauer: Fixed in 70cd5e8

typhos/jira.py
ZLLentz: Oh, we probably also want the `PYTHONPATH` environment variable? Or is that too much?

ZLLentz (approved):
I think this is really snazzy

typhos/jira.py
klauer: Couldn't hurt... probably! I guess we'll see if there's a limit to the length the issue collector will accept.

klauer: Adding that long list necessitated some formatting tweaks, which ended up making this better overall. Viewed from Jira:
image

slacAWallace:
Game changer!

PR #446, opened 2021-07-26
API: move pcds templates to different repo

ZLLentz:
## Description

These have been shipped over to pcdsdevices, which is where the classes are defined. As such, these are being removed here.

## Motivation and Context

Do not want to duplicate templates in two places, this will become confusing.

## How Has This Been Tested?

It will be tested on the other end, where the templates are moving to. This might break some of the local tests which will need to be adjusted.

A note about the moved templates has been added to the templates page.


ZLLentz:
`pcdsdevices` needs a tag on `typhos` to merge the pr that would allow the docs link here to work. We're chicken-and-egged. But I think merging this one first is better.

klauer (approved):
LGTM

PR #447, opened 2021-07-30
ENH: Expand positioner widget, add entry points, other tweaks

ZLLentz:
## Description

Add the following features to the positioner widget:
- Error message from the IOC
- Moving/done moving indicator
- Expert screen button (optional, opens up a full typhos suite)
- Alarm indicator
- Clear error button (if clear_error method is defined)
- Remove the stop button if stop method is overridden with None
- Various aesthetic tweaks
- Fix bugs related to the combobox version selecting when it isn't supposed to, and not selecting when it is supposed to
- More misc bugfixes

Motor record version example (plus custom error message + clearing)
image

Combobox version example:
image

Add the `typhos.ui` entry point:
- Must refer to a `pathlib.Path`, `str`, or `list` of those types of objects
- These are paths that we'll check for `typhos` templates

Misc:
- Add an examples submodule for launching example screens (e.g. `python -m typhos.examples.positioner`)
- Modify the alarm widgets to allow customizable pen width, pen color, and pen style for the border. This is used in the positioner widget to make the indicator fit in better
- Fix an obscure issue with importing `QWebEngineView` inside Qt `designer`
- Fix an issue where the related display module added a happi requirement. This is supposed to be an optional dependency.

## Motivation and Context

- I got a call in the middle of my weekend and I want to not get that particular call, which should not have happened with better ui here.
- Entry points are a good way to refer to resources in other python repos

## How Has This Been Tested?

Interactively, with the example screens, and some unit tests added for cases where a failure will be visually confusing to the user.

Added docs page.


ZLLentz:
I need to actually build and check the docs locally and then this is done (ran out of time today)

typhos/positioner.py
```diff
@@ -194,6 +219,30 @@ def stop(self):
...
+
+        The device may have errors in the IOC. These will be cleared by calling
+        the clear_error method.
+
+        The screen may have errors from the status of the last move. These will
+        be cleared from view.
+        """
+        for device in self.devices:
+            try:
+                device.clear_error()
```
klauer, line 235: Thinking that there may be a `caput` behind `clear_error` that may block the GUI thread, I wonder if this should be done in the background?

typhos/positioner.py
```diff
@@ -255,6 +304,31 @@ def _link_limits_by_limits_attr(self):
         self.ui.low_limit.hide()
         self.ui.high_limit.hide()
 
+    @utils.linked_attribute('moving_attribute', 'ui.moving_indicator', True)
+    def _link_moving(self, signal, widget):
+        """Link the positioner moving indicator with the ui element."""
+        if signal is None:
+            widget.hide()
+            return False
+        else:
```
klauer, line 313: Nitpicky optional style suggestion - remove `else:` and de-indent the following block

typhos/positioner.py
```diff
@@ -269,7 +343,12 @@ def _define_setpoint_widget(self):
         if selection:
             self.ui.set_value = QtWidgets.QComboBox()
             self.ui.set_value.addItems(setpoint_signal.enum_strs)
-            self.ui.set_value.currentIndexChanged.connect(self.set)
+            self.ui.set_value.activated.connect(self.set)
```
klauer, line 346: A small but very important fix, changing the `set` signal to `activated`.

typhos/positioner.py
```diff
@@ -298,9 +377,31 @@ def add_device(self, device):
         self._link_low_limit_switch()
         self._link_high_limit_switch()
 
+        try:
+            if device.stop is None:
```
klauer, line 381: `ophyd.Device` is guaranteed to have `stop` as an attribute. The guard below against `AttributeError` feels appropriate to me regardless.

typhos/utils.py
```diff
@@ -73,12 +82,47 @@
...
+            entry_objs.append(obj)
+
+    for obj in entry_objs:
+        try:
+            yield pathlib.Path(obj)
+        except Exception:
+            msg = (f'{TYPHOS_ENTRY_POINT_KEY} entry point '
+                   f'{entry.name}: {obj} is not a valid path!')
+            logger.error(msg)
+            logger.debug(msg, exc_info=True)
```
klauer, line 124: Random thought, but remember the `stacklevel` thing I was talking about the other day?

I think this is a pretty good example of where it'd be useful. We can now have:
`utils.log_error_and_debug_exception_message(logger, msg)` 
with `logger.error(msg, stacklevel=2)` and `logger.debug(msg, exc_info=True, stacklevel=2)` would indicate that the message came from this function.

That said, I think we're still 3.6+ on typhos, so ... a note for the future?

klauer (approved):
This is a big improvement for positioners which get a lot of usage in and out of our group.

Interactive testing worked really nicely, docs look great, tests look great. 👍 Some general comments, nothing beyond a bit of nitpicking I think. None necessary to accept.

docs/source/conf.py
```diff
@@ -186,7 +189,8 @@
 ]
 
 
-
-
 # Example configuration for intersphinx: refer to the Python standard library.
-intersphinx_mapping = {'https://docs.python.org/': None}
+intersphinx_mapping = {
+    'pydm': ('https://slaclab.github.io/pydm-tutorial/', None),
```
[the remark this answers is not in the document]
ZLLentz: I was surprised they were on different doc sites while I was setting this up. I agree, we'll add it as soon as we want to link something from the main docs.

docs/source/basic_usage.rst
```diff
@@ -238,3 +238,10 @@ and a pre-configured issue collector.
    ``"Bearer ${TYPHOS_JIRA_TOKEN}"``.
 4. ``TYPHOS_JIRA_EMAIL_SUFFIX`` (str): the default e-mail suffix to put on
    usernames, such as ``"@example.com"``.
+
+
+Launching the Examples
+======================
+There are example screens in the ``typhos.examples`` submodule. After
+installing ``typhos``, you can launch them using e.g.
+``python -m typhos.examples.positioner``.
```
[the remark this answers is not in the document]
ZLLentz: Do you think I should list them here? I think I should, I was just a little bit lazy when writing the docs since updating the docs when more are added is pain.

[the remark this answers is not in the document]
klauer: I think it'd be worthwhile to. Copy/paste is perfect for lazy users like me.

typhos/positioner.py
ZLLentz: Good catch, this will freeze the gui if the network is being slow

typhos/positioner.py
ZLLentz: Ah, right, when I wrote this the first time I didn't have a `return` statement in the first if section, so originally it needed to be like this.

Minimizing indent levels in Python is actually a pretty important thing... Easy to get into trouble if you aren't diligent about it.

typhos/positioner.py
ZLLentz: Should I point this out with a descriptive comment, or is that overboard?

typhos/positioner.py
ZLLentz: Paranoia as a policy of course

typhos/positioner.py
```diff
@@ -315,7 +416,31 @@ def moving(self):
     def moving(self, value):
         if value != self._moving:
             self._moving = value
-            utils.reload_widget_stylesheet(self, cascade=True)
+            self._after_set_moving(value)
+
+    def _after_set_moving(self, value):
+        """
+        Common updates needed after a change the moving state.
```
ZLLentz, line 423:
```suggestion
        Common updates needed after a change to the moving state.
```

typhos/positioner.py
ZLLentz: I'll go with the mantra of "If you think it might warrant a comment, and you can't decide, just make the comment"

klauer: Sounds good to me 😁

docs/source/basic_usage.rst
```diff
@@ -238,3 +238,12 @@ and a pre-configured issue collector.
...
 4. ``TYPHOS_JIRA_EMAIL_SUFFIX`` (str): the default e-mail suffix to put on
    usernames, such as ``"@example.com"``.
+
+
+Launching the Examples
+======================
+There are example screens in the ``typhos.examples`` submodule. After
+installing ``typhos``, you can launch them as follows:
+
+- ``python -m typhos.examples.panel``.
```
klauer, line 248:
```suggestion
- ``python -m typhos.examples.panel``
```

docs/source/basic_usage.rst
```diff
@@ -238,3 +238,12 @@ and a pre-configured issue collector.
...
+- ``python -m typhos.examples.positioner``.
```
klauer, line 249:
```suggestion
- ``python -m typhos.examples.positioner``
```

docs/source/basic_usage.rst
ZLLentz: Even if it doesn't let's destroy it

ZLLentz:
Ok let's ship it

PR #448, opened 2021-08-05
DOC: v2.0.0 rel notes

ZLLentz:
## Description

Let us tag our new features!
There was a lot of progress in the last month!

Goal is to tag and release tomorrow, Thursday 8/5

## Motivation and Context

I want to use these!

## How Has This Been Tested?

I built the docs locally to verify that the release notes look normal.

This is docs


klauer (approved):
I'm excited to try these features, too. (Now if only confluence would get that upgrade we need...)

Release notes look good. `2.0.0` 👏

ZLLentz:
Ah I missed one feature line here- adding and then merging

ZLLentz:
This is going to be delayed for a bit until we work out a revision to the "removing stop button from the positioner widget" api

PR #449, opened 2021-08-05
MNT: hide stop only on attribute error to avoid bluesky trap

ZLLentz:
## Description

Remove the option to set stop to None
Require stop to raise AttributeError to remove the button
Get it in before the v2.0.0 tag
In the last tag: the stop button would appear no matter what, so this won't be an API break except on master.

## Motivation and Context

Bluesky logs at the exception level if a call to stop() fails, as it should.
The API here for typhos should not encourage anyone to get into that state.
It's also really messy to override a function with None.

## How Has This Been Tested?

Interactively

Need to add a docs page suggesting how to override with attribute error


klauer (approved):
LGTM

PR #450, opened 2021-08-09
ENH: add suite options for layouts, display types, scrollbars, and starting window size

ZLLentz:
## Description

Add the following options to the CLI and their corresponding handlers in the library:
- `--layout`: pick between horizontal (default), vertical, grid, or flow layouts. The flow layout is lifted from PyDM and allows the user to resize the window and have the widgets populate the space as needed, e.g. a semi-dynamic grid.
- `--cols`: select how many columns the grid layout has
- `--display-type`: pick between embedded, detailed (default), or engineering screen as the kind of display to open at the start.
- `--scrollable`: pick between auto (default), true, and false to control whether the screen can have scrollbars or not. While auto is the default, this is actually new behavior: allows scrollbars for detailed and engineering screens, but not for embedded screens.
- `--size`: pick a starting size for the window. The default is whatever the window's size hint is (I think?)

## Motivation and Context

- TMO has been using a horizontal layout of ~17 detailed screens in a suite which takes up a ton of horizontal space and needs some clicking around to make them all embedded.
- In general, some options for how to arrange these gives us ways to quickly generate reasonable groups of screens.
- The grid and flow layouts are probably most appropriate for showing lots of embedded screens.
- The vertical layout may not be so useful, but it felt weird to omit it.
- Some screens have users clicking into the embedded mode for all elements, it would be better if they started in the display type that the user wanted.
- Scrollbars can render strangely in the grid layout and for embedded widgets, so I added some options to handle these cases.
- Some windows (flow layouts in particular) open at strange sizes, presumably because there isn't a good way to guess what the user is going to want. For a suite launcher script, being able to specify a starting size improves convenience.

## How Has This Been Tested?

Only interactively so far, but it did not break existing tests.

Only in docstrings and in the `--help` text. There isn't a good user-facing page in the docs right now for the CLI.

See comments for the most up-to-date screenshots.

### DIFF
--- a/typhos/cli.py
+++ b/typhos/cli.py
@@ -17,7 +17,7 @@ import typhos
 from typhos.app import get_qapp, launch_suite
 from typhos.benchmark.cases import run_benchmarks
 from typhos.benchmark.profile import profiler_context
-from typhos.display import DisplayType
+from typhos.display import DisplayTypes
 from typhos.utils import nullcontext
 
 logger = logging.getLogger(__name__)
@@ -177,11 +177,11 @@ class FixedColGrid(QtWidgets.QGridLayout):
 
 def get_display_type_from_cli(display_type):
     if 'embedded'.startswith(display_type):
-        return DisplayType.embedded_screen
+        return DisplayTypes.embedded_screen
     if 'detailed'.startswith(display_type):
-        return DisplayType.detailed_screen
+        return DisplayTypes.detailed_screen
     if 'engineering'.startswith(display_type):
-        return DisplayType.engineering_screen
+        return DisplayTypes.engineering_screen
     else:
         raise ValueError(
             f'{display_type} is not a valid display type. '
@@ -300,7 +300,7 @@ def typhos_cli(args):
                                cfg=args.happi_cfg,
                                fake_devices=args.fake_device,
                                layout=args.layout,
-                               cols=args.cols,
+                               cols=int(args.cols),
                                display_type=args.display_type,
                                )
         return suite
--- a/typhos/cli.py
+++ b/typhos/cli.py
@@ -10,11 +10,14 @@ import sys
 import coloredlogs
 import pcdsutils
 from ophyd.sim import clear_fake_device, make_fake_device
+from pydm.widgets.template_repeater import FlowLayout
+from qtpy import QtWidgets
 
 import typhos
 from typhos.app import get_qapp, launch_suite
 from typhos.benchmark.cases import run_benchmarks
 from typhos.benchmark.profile import profiler_context
+from typhos.display import DisplayType
 from typhos.utils import nullcontext
 
 logger = logging.getLogger(__name__)
@@ -28,10 +31,20 @@ parser.add_argument('devices', nargs='*',
                     help='Device names to load in the TyphosSuite or '
                          'class name with parameters on the format: '
                          'package.ClassName[{"param1":"val1",...}]')
-parser.add_argument('--embed', action='store_true',
-                    help='Load the embedded templates instead of the '
-                         'detailed templates. Good for suites with many '
-                         'devices.')
+parser.add_argument('--layout', default='horizontal',
+                    help='Select a alternate layout for suites of many '
+                         'devices. Valid options are "horizontal" (default), '
+                         '"vertical", "grid", "flow", and any unique '
+                         'shortenings of those options.')
+parser.add_argument('--cols', default=3,
+                    help='The number of columns to use for the grid layout '
+                         'if selected in the layout argument. This will have '
+                         'no effect for other layouts.')
+parser.add_argument('--display_type', default='detailed',
+                    help='The kind of display to open for each device at '
+                         'initial load. Valid options are "embedded", '
+                         '"detailed" (default), "engineering", and any '
+                         'unique shortenings of those options.')
 parser.add_argument('--happi-cfg',
                     help='Location of happi configuration file '
                          'if not specified by $HAPPI_CFG environment variable')
@@ -111,7 +124,8 @@ def _create_happi_client(cfg):
     return happi.Client.from_config(cfg=cfg)
 
 
-def create_suite(device_names, cfg=None, fake_devices=False, embedded=False):
+def create_suite(device_names, cfg=None, fake_devices=False,
+                 layout='horizontal', cols=3, display_type='detailed'):
     """Create a TyphosSuite from a list of device names."""
     if device_names:
         devices = create_devices(device_names, cfg=cfg,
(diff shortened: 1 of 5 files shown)


ZLLentz:
Setting some goals here:
1. Allow the user to decide which template to use and which layout to use from the cli, as well as in `TyphosSuite.from_devices`
2. Fix the sizing/whitespace issues with suites that have many small devices: in general, make the screens load at minimum size. Perhaps another cli/from_devices switch for sizing. The issue here is related to how we set the `sizeHint` in display.py, but it's not clear to me what the best resolution is yet. It's also not clear to me at which point the suite should ask the display for scrollbars- maybe this should also be configurable? I'm not sure. The problem is that large displays need scrollbars and certain small displays collapse to nothing without them.

In a different PR, I'll look into fixing the layout issues for suites of 1 large device- it really should just look like a typhos suite.

Current progress:
image

ZLLentz:
I think this is now, functionally, as far as I plan to take it, or at least close. The last thing I'm considering is a cli option to set the initial window size, so something like the flow layout can open at a sensible starting size by default. There is still code cleanup to be done before asking for a review.

Here are a few examples of the previous example screen:

- Awkward pre-PR layout `typhos --fake-device`

image


- Using the new grid layout + start embedded: `typhos --layout grid --cols 4 --display-type emb --fake-device`

image


- Using the new flow layout + start embedded (and resizing a bit): `typhos --layout flow --display-type emb --fake-device`

image

ZLLentz:
The only thing that still bothers me is that the flow layout's row height is locked to be even with the tallest row, adding some extra whitespace to the screen.

typhos/cli.py
```diff
@@ -6,60 +6,130 @@
...
-                         'If omitted, prints results to stdout. '
-                         'Turns on line profiling.')
-parser.add_argument('--benchmark', nargs='*',
-                    help='Runs the specified benchmarking tests instead of '
-                         'launching a screen. '
-                         'If no specific tests are specified, '
-                         'runs all of them. '
-                         'Turns on line profiling.')
+parser = argparse.ArgumentParser(
+    description='Create a TyphosSuite for '
```
klauer, line 29: This feels medium-high-level nitpicky, but I think there is room for improvement. Consider this instead:

```python
parser.add_argument(
    'devices',
    nargs='*',
    help=(
        "Device names to load in the TyphosSuite or class name with "
        "parameters of the format: package.ClassName[{"param1":"val1",...}]."
    ),
)
```

Primarily: `(` + newline + strings at same indent level, and final closing-line ")"

typhos/cli.py
```diff
@@ -107,15 +178,162 @@ def _create_happi_client(cfg):
...
+        String representation of the layout.
+    cols : int
+        Number of columns to use for the grid layout.
+
+    Returns
+    -------
+    qlayout : QLayout
+        The qt layout object, instantiated.
+    """
+    if 'horizontal'.startswith(layout):
```
klauer, line 259: `.lower()` for these, or is case-sensitivity OK?

typhos/cli.py
```diff
@@ -107,15 +178,162 @@ def _create_happi_client(cfg):
...
+    ):
+        row = self._widget_index // self._n_cols
+        col = self._widget_index % self._n_cols
+        super().addWidget(widget, row, col, *args)
+        self._widget_index += 1
+
+
+def get_display_type_from_cli(display_type: str) -> DisplayTypes:
+    """Convert the cli string to the appropriate DisplayTypes enum."""
+    if 'embedded'.startswith(display_type):
```
klauer, line 309: Would be kinda nice to do this directly on the enum, but then again it probably won't be reused outside of the CLI, so :meh:

typhos/cli.py
```diff
@@ -107,15 +178,162 @@ def _create_happi_client(cfg):
...
+        return DisplayTypes.embedded_screen
+    if 'detailed'.startswith(display_type):
+        return DisplayTypes.detailed_screen
+    if 'engineering'.startswith(display_type):
+        return DisplayTypes.engineering_screen
+    else:
```
klauer, line 315: `else` unnecessary in these; can save you a few bytes o' indentation.

typhos/display.py
```diff
@@ -74,6 +90,37 @@ def normalize_display_type(display_type):
...
+
+    Raises
+    ------
+    ValueError
+        If the input cannot be made a :class:`ScrollOptions`.
+    """
+    try:
+        return ScrollOptions(scroll_option)
+    except Exception as ex:
+        if scroll_option in ScrollOptions.names:
```
klauer, line 117: I think a shorter way might be (1) try `ScrollOptions(scroll_option)` (2) try `ScrollOptions[scroll_option]` (note brackets)

typhos/display.py
```diff
@@ -1437,7 +1517,7 @@ def from_class(cls, klass, *, template=None, macros=None, **kwargs):
             obj = pcdsutils.utils.get_instance_by_name(klass, **kwargs)
         except Exception:
             logger.exception('Failed to generate TyphosDeviceDisplay from '
-                             'device %s', obj)
+                             'class %s', klass)
```
klauer, line 1520: Seemingly minor fix, but catches `NameError` failure - nice!

klauer (approved):
I can't wait to give this a spin. Only minor suggestions and nitpicky ones at that. I'd say merge at will.

typhos/cli.py
ZLLentz: You're definitely right here. I thought I was being clever by just running `black` and hoping for the best, but I didn't think much about the formatting afterwards.

typhos/cli.py
ZLLentz: I think you're right, we should be kind to the cli user since it will be unambiguous

typhos/cli.py
ZLLentz: Is it worth putting an enum base in pcdsutils for this? This should be achievable generically...

typhos/cli.py
ZLLentz: you're right

typhos/display.py
ZLLentz: You are correct, this is a copy/paste/nothink from a similar function above for the display types. I'll update both of them.

typhos/display.py
ZLLentz: VSCode pointed it out to me, this must have been lurking in the code for a while

ZLLentz: I think this is an easy one to miss without a static analysis tool since the `obj` name is right there so it seems ok to use when it isn't

typhos/cli.py
klauer: It's generally useful for command-line input, so maybe?

ZLLentz:
Ok I did the things let's go

PR #458, opened 2021-10-14
ENH: add option to pop out documentation frame

klauer:
## Description
Open docs in a new window, without resorting to opening in a browser.

## Motivation and Context
We are currently hopeful the built-in documentation will be of general use.
However, our screens generally have an aspect ratio that's taller than it is wide, which is poor for viewing Confluence.
Looking at the screenshot below, you can easily pop out the help documentation frame into a new window, or pop it back in if needed.

## How Has This Been Tested?
Interactively

This PR text

image

### DIFF
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -470,6 +470,9 @@ class TyphosHelpToggleButton(TyphosToolButton):
 
     Attributes
     ----------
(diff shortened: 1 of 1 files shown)


ZLLentz:
I'll merge this now

PR #463, opened 2021-11-18
REF: direct-to-PV vs signal protocol by way of cl

klauer:
## Description
Typhos supports two paths to get data from ophyd Signals.
1. Peeking into the `EpicsSignal` to see which PVs it is using and passing that along to PyDM
2. Directly through the ophyd subscription mechanism

This PR allows for an escape hatch for (1) in scenarios where we want typhos to use method (2).
This escape hatch comes in the form of a control layer `cl` check to see if the user has customized it beyond what ophyd supports.

## Motivation and Context
Related to https://github.com/pcdshub/atef/pull/17
Where we want archived data to be sourced rather than live data

## How Has This Been Tested?
Interactively, so far

This PR text.

This is a typhos screen using this PR that gets data from the EPICS archiver:
image

### DIFF
--- a/typhos/widgets.py
+++ b/typhos/widgets.py
@@ -1042,6 +1042,9 @@ def _get_ndimensional_widget_class(dimensions, desc, variety_md, read_only):
     }.get(dimensions, TyphosLabel)
 
 
+DIRECT_CONTROL_LAYERS = {"pyepics", "caproto"}
+
+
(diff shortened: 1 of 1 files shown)


ZLLentz (approved):
This is much simpler than I thought it would be. I didn't notice that the control layers had `name` attributes.

PR #464, opened 2021-11-19
MNT: fix newly found failure in get_native_methods

ZLLentz:
## Description

- Fix the CI to run two tests that were accidentally skipped
- Bring the `set.add` call into the `TypeError` try/except block that is intended to catch us trying to add an unhashable type to the set.
- Make the application handling in test_cli more consistent to avoid a hanging test suite

I don't know why this was working before or is failing now. The new object that passed the first check and made it to the add method was a `dict[str, str]`

## Motivation and Context

Two tests were failing silently because there were accidentally skipped.

## How Has This Been Tested?

Unit tests pass, change seems innocuous

Will be in the release notes

### DIFF
--- a/typhos/tests/conftest.py
+++ b/typhos/tests/conftest.py
@@ -6,6 +6,7 @@ from functools import wraps
 
 import numpy as np
 import ophyd.sim
+import pydm
 import pytest
 import qtpy
 from happi import Client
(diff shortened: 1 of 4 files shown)


ZLLentz:
Hey, it worked!
Not sure whether to merge and fix in a separate PR or to continue working here to investigate the issue

ZLLentz:
I'll keep working here to resolve this and keep the master ci green

klauer (approved):
Good catch. LGTM 👍

ZLLentz:
Ok, the problem in #465 might actually be a consistent thing on py39- it has happened twice in a row on travis linux and always happens on my windows. I'll try to run the travis a third time and if it still hangs I'll need to solve that before merging this.

ZLLentz:
We've reached the terrible point where I've solved the freeze on my local machine but travis is still hanging. I'll have to move my testing to our linux servers to see if I can reproduce it there.

ZLLentz:
I can no longer reproduce the freezing at all on any computer I have access to. It only happens on python 3.9 which I find very strange. I'm not sure what to do except start working on other CI items in the issue tracker and hope they are related.

ZLLentz:
Ok that last try seemed to do the trick thank goodness

typhos/tests/conftest.py
```diff
@@ -63,6 +64,15 @@ def qapp(pytestconfig):
     return application
 
 
+@pytest.fixture(scope='function', autouse=True)
```
klauer, line 67: Nicely done!

PR #472, opened 2022-01-31
BUG: fix issue where combobox positioners can dangerously trigger moves via mousewheel

ZLLentz:
## Description

Follow-up to #389 
Copy that fix over to the combo boxes used for the positioner widget so that when you scroll over the motor destination you don't accidentally trigger a move, e.g. if you are scrolling up and down a long screen.

## Motivation and Context

This causes unintended moves which can and do lead to lost beam time.

## How Has This Been Tested?

Locally with no write access to the real PVs, comparing before and after the patch.

N/A


klauer (approved):
I see - so the other was our customized PyDMEnumComboBox, whereas this is the one for positioners. Both definitely need the fix.

Looks good to me!

PR #475, opened 2022-02-04
FIX: write pv assumption

### LINKED ISSUE #473: Assumption of _write_pv may not be accurate for all signal types
(More details to follow; and hopefully a fix!)

## Current Behavior
```python
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.1.1/lib/python3.9/site-packages/typhos/utils.py", line 1319, in wrapped
    link_signal_to_widget(signal, widget)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.1.1/lib/python3.9/site-packages/typhos/utils.py", line 1274, in link_signal_to_widget
    widget.channel = channel_from_signal(signal, read=read)
  File "/cds/group/pcds/pyps/conda/py39/envs/pcds-5.1.1/lib/python3.9/site-packages/typhos/utils.py", line 161, in channel_from_signal
    return channel_name(signal._write_pv.pvname)
AttributeError: 'PytmcSignalRO' object has no attribute '_write_pv'
```

1.
2.
3.

### LINKED ISSUE #474: Typo in typhos.positioner "count" -> "could"
https://github.com/pcdshub/typhos/blob/39ac7c8a066008cd8fd4ac01e1a23db08eb9ee81/typhos/positioner.py#L647

klauer:
## Description
`_write_pv` and `_read_pv` are generally good assumptions for `EpicsSignalBase` and its subclasses. However, typhos should be lenient if/when access to these attributes fails.

pcdsdevices may also require adjustments to `PytmcSignalRO`, but that shouldn't affect this PR.

## Motivation and Context
Closes #474 
Closes #473 

## How Has This Been Tested?
Test suite and interactively with problematic `PytmcSignalRO` signal.

Issue and PR text.

### DIFF
--- a/typhos/positioner.py
+++ b/typhos/positioner.py
@@ -644,7 +644,7 @@ def clear_error_in_background(device):
         except AttributeError:
             pass
         except Exception:
-            msg = 'Count not clear error!'
+            msg = "Could not clear error!"
             logger.error(msg)
             logger.debug(msg, exc_info=True)
 
--- a/typhos/utils.py
+++ b/typhos/utils.py
@@ -154,11 +154,23 @@ def channel_from_signal(signal, read=True):
     """
     Create a PyDM address from arbitrary signal type
     """
-    # Add an item
     if isinstance(signal, EpicsSignalBase):
         if read:
-            return channel_name(signal._read_pv.pvname)
-        return channel_name(signal._write_pv.pvname)
+            # For readback widgets, focus on the _read_pv only:
+            attrs = ["_read_pv"]
+        else:
+            # For setpoint widgets, _write_pv may exist (and differ) from
+            # _read_pv, try it first:
+            attrs = ["_write_pv", "_read_pv"]
+
+        # Some customizations of EpicsSignalBase may have different attributes.
+        # Try the attributes, but don't fail if they are not present:
+        for attr in attrs:
+            pv_instance = getattr(signal, attr, None)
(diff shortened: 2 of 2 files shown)


ZLLentz:
I think this looks correct to me, and should work for both the pyepics and caproto backends.

Do you remember exactly which device was throwing this error so we can test it? If not, I can start opening all the screens again.

klauer:
I don't recall the device, but it had a `PytmcSignalRO` if I recall correctly, so that should narrow it down.

ZLLentz:
That was enough to find it: the devices are `rtdsl0` and `rtdsk0`, whose subdevices have a `state` setpoint that is a `PytmcSignalRO`... so they don't function as movable state devices. I think they need an additional fix in `pcdsdevices` but this probably removes the terminal spam error (I'll check)

ZLLentz:
The aforementioned screens are very happy with this PR. We can check if they would actually move their respective devices over the PAMM but I think this should be merged.

klauer:
Nice - OK! 👍 For what it's worth, this was totally untested...

tangkong (approved):
Looks more robust to me, screens also load without protest.

ZLLentz:
ah right I'll just make sure the filters move on the screen after lunch and then press merge

ZLLentz:
OK the device on the pcdsdevices side definitely needs a looking at but this is pretty good for all normal devices and it removes the error

PR #477, opened 2022-02-08
ENH: add option to hide displays in suite at start

ZLLentz:
## Description

Add `--hide-displays` option to open a typhos suite with all displays hidden.

image

## Motivation and Context

If there are too many devices, the typhos suite is hard to use when everything opens immediately.

## How Has This Been Tested?

Interactively only.

Not yet

### DIFF
--- a/typhos/cli.py
+++ b/typhos/cli.py
@@ -88,6 +88,14 @@ parser.add_argument(
     ),
 )
 parser.add_argument(
+    '--hide-displays',
+    action='store_true',
+    help=(
+        'Option to start with subdisplays hidden instead '
+        'of shown.'
+    )
+)
+parser.add_argument(
     '--happi-cfg',
     help=(
         'Location of happi configuration file '
@@ -207,6 +215,7 @@ def create_suite(
     cols: int = 3,
     display_type: str = 'detailed',
     scroll_option: str = 'auto',
+    show_displays: bool = True,
 ) -> TyphosSuite:
     """
     Create a TyphosSuite from a list of device names.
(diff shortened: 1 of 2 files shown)


typhos/cli.py
```diff
@@ -87,6 +87,14 @@
         'your application. Example: --size 1000,1000'
     ),
 )
+parser.add_argument(
+    '--hide-displays',
```
klauer, line 91: Shortcut?
```suggestion
    "-h", "--hide-displays",
```

or maybe even
```suggestion
    "-M", "--minimal",
    dest="hide_displays",
```

klauer (approved):
Looks good to me! Just a thought on the CLI arg, which at least deserves a short form.

typhos/cli.py
ZLLentz: I think making the shortcut `-h` shadows the `-h`/`--help` and should be avoided.
Maybe just `--hide`? I'm not sold on `-m`/`--minimal`

klauer: Just throwing out ideas, I'm not particularly fond of any of them, so the right answer is probably the lazy answer.

tangkong (approved):
looks good to me.  I'm somewhat surprised this wasn't the default before

typhos/cli.py
ZLLentz: And I'm throwing ideas back! It's a dialog. I'm also leaning toward lazy at this time.

klauer: Unintentionally dismissive/salty sounding reply above, sorry! Thinking back, I still can't come up with anything better. I'd say this is good to go.

PR #478, opened 2022-02-09
Allow use of sig:// protocol in typhos templates

ZLLentz:
## Description

Register a device's signals when it is added to a `TyphosDeviceDisplay` so that device templates can include channels like `sig://${name}.setpoint` that refer to named signals on the device.

Fix a bug where suites could not find subdisplays for certain devices with underscores in their names in certain situations.

## Motivation and Context

We need to make typhos templates easier to construct from device definitions and more flexible for non-EPICS devices.

## How Has This Been Tested?

Added some tests
Needs interactive testing before merge

Not yet

### DIFF
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -21,6 +21,7 @@ from . import cache
 from . import panel as typhos_panel
 from . import utils, web, widgets
 from .jira import TyphosJiraIssueWidget
+from .plugins.core import register_signal
 
 logger = logging.getLogger(__name__)
 
@@ -1362,6 +1363,10 @@ class TyphosDeviceDisplay(utils.TyphosBase, widgets.TyphosDesignerMixin,
            3. The argument ``macros`` is then used to fill/update the final
               macro dictionary.
 
+        This will also register the device's signals in the sig:// plugin.
+        This means that any templates can refer to their device's signals by
+        name.
+
         Parameters
         ----------
         device : ophyd.Device
@@ -1376,6 +1381,9 @@ class TyphosDeviceDisplay(utils.TyphosBase, widgets.TyphosDesignerMixin,
             self.devices.clear()
         # Add the device to the cache
         super().add_device(device)
+        logger.debug("Registering signals from device %s", device.name)
+        for component_walk in device.walk_signals():
+            register_signal(component_walk.item)
         self._searched = False
         self.macros = self._build_macros_from_device(device, macros=macros)
         self.load_best_template()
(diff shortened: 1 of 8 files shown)


typhos/tests/conftest.py
```diff
@@ -230,7 +230,7 @@ def hints(self):
 
 @pytest.fixture(scope='function')
 def device():
-    dev = MockDevice('Tst:This', name='Simulated Device')
+    dev = MockDevice('Tst:This', name='simulated_device')
```
ZLLentz, line 233: It turns out that this name change is what breaks the suite tests. I'm a bit baffled, will look into it closer- is there a latent typhos bug that has been chilling?
Hopefully not... It turns out I didn't NEED to change this- it was part of my debug process as something that looked wrong.

ZLLentz: I found and fixed the related bug in https://github.com/pcdshub/typhos/pull/478/commits/6970d70f05edc481baac2931777c159f7996413b

It turns out that we store the "cleaned up" name only, so when I made the device have a "normal" name it was unable to find the panel.

klauer: Great find!

typhos/display.py
```diff
@@ -1376,6 +1381,9 @@ def add_device(self, device, macros=None):
             self.devices.clear()
         # Add the device to the cache
         super().add_device(device)
+        logger.debug("Registering signals from device %s", device.name)
+        for component_walk in device.walk_signals():
+            register_signal(component_walk.item)
```
klauer, line 1386: Probably the biggest impact for the general case is this loop here. I imagine it's not bad performance-wise, though.

typhos/plugins/core.py
```diff
@@ -21,14 +20,41 @@ def register_signal(signal):
...
-    signal_registry[signal.name] = signal
+    for name in names:
+        if name in signal_registry:
+            # Case 1: harmless re-add
+            if signal_registry[name] is signal:
+                logger.debug(
+                    "The signal named %s is already registered!",
+                    name,
+                )
+            # Case 2: harmful overwrite! Name collision!
```
klauer, line 48: Good call!

klauer (approved):
Looks good to me. Great QOL improvement for custom screen developers.

typhos/display.py
ZLLentz: It shouldn't matter too much, but there was probably an alternate path where we could stash the devices and have the sig:// plugin find the signal at runtime, a fast process if we use the dotted names.

PR #480, opened 2022-02-24
BUG: Workaround for C++ Wrapped Exception

### LINKED ISSUE #470: Exception: wrapped C/C++ object of type TyphosAlarmCircle has been deleted
## Expected Behavior
No error.

## Current Behavior
Repetitive errors being caught by logging system (hundreds over a month):
```
  File "/cds/group/pcds/pyps/conda/py39/envs/dev/bin/lucid", line 10, in <module>
    sys.exit(main())
  File "/cds/group/pcds/pyps/apps/dev/pythonpath/lucid/launcher.py", line 246, in main
    launch(**kwargs)
  File "/cds/group/pcds/pyps/apps/dev/pythonpath/lucid/launcher.py", line 240, in launch
    app.exec_()
  File "/cds/group/pcds/pyps/apps/dev/pythonpath/typhos/alarm.py", line 301, in update_severity
    self.update_current_alarm()
  File "/cds/group/pcds/pyps/apps/dev/pythonpath/typhos/alarm.py", line 316, in update_current_alarm
    self.alarm_changed.emit(new_alarm)
RuntimeError: wrapped C/C++ object of type TyphosAlarmCircle has been deleted
```

I have not tracked this to the source just yet, whether it's due to LUCID's usage or the typhos internals.

## Possible Solution
Track down issue here or in LUCID where this occurs.

Unclear at this point; no direct user reports.

## Context
Seen in log messages from LUCID home screens.

## Your Environment
Seen messages from pcds-4.2.0, pcds-dev, etc. I believe it's still an issue.

ZLLentz:
## Description

Simply ignore this error when it pops up.
For background on why this happens, see https://github.com/slaclab/pydm/issues/798

If we do fix how the cleanup happens here, this PR is still needed because it's possible for a race condition to hit this line at a bad time anyway.

I snuck a side change in about setting up the devices list earlier because it helped me in the debugging process by fixing some of the reprs during the initial load.

## Motivation and Context

closes #470 

## How Has This Been Tested?

Interactively only

It'll be in the release notes

### DIFF
--- a/typhos/alarm.py
+++ b/typhos/alarm.py
@@ -313,11 +313,18 @@ class TyphosAlarm(TyphosObject, PyDMDrawing, _KindLevel, _AlarmLevel):
         else:
             new_alarm = max(info.alarm for info in self.signal_info.values())
         if new_alarm != self.alarm_summary:
-            self.alarm_changed.emit(new_alarm)
-            logger.debug(
-                f'Updated alarm from {self.alarm_summary} to {new_alarm} '
-                f'on alarm widget with channel {self.channels()[0]}'
+            try:
+                self.alarm_changed.emit(new_alarm)
+            except RuntimeError:
+                # Widget was destroyed and not properly cleaned up
+                logger.debug('Dangling reference to alarm widget!')
+                return
+            else:
+                logger.debug(
+                    f'Updated alarm from {self.alarm_summary} to {new_alarm} '
+                    f'on alarm widget with channel {self.channels()[0]}'
                 )
+
         self.alarm_summary = new_alarm
 
     def set_alarm_color(self, alarm_level):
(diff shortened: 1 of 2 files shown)


klauer (approved):
Great work digging into this. This one's bugged me for a while, and I wasn't able to get at the real source of it.

This patch looks good to me, and I'm hopeful we can avoid the _non-race condition version_ of this with pydm fixes upstream.

ZLLentz:
Thanks Ken! I'd like to add briefly that it seems there is no simple way to patch this behavior on a widget-by-widget basis if the widget has `PyDMWidget` as a parent. Once the first `destroying` signal is emitted, it reaches the non-disconnecting destroy block and then forgets that the channel ever existed, preventing any additional future handling without some extreme monkeypatching or poking at the pydm internals.

tangkong (approved):
LGTM, I'll not hold this up as I learn more about typhos

Issue #484, opened 2022-04-04
Toggling visibility via clicking title inconsistent in behavior


klauer:
#### Option 1:
Clicking a title should hide or show all sub-items in a `detailed_tree`.

#### Option 2:
Or alternatively, if this behavior is undesired/not useful, it should be removed.

## Current Behavior
This works sometimes but not always due to the following reasons:
1. https://github.com/pcdshub/typhos/blob/408c10892f25398b9e79f7cc9286c273ac918b6e/typhos/display.py#L1600
    May have some widgets that are visible and some that are not, and checking the status of `all()` can result in the same `False` value each time. Toggling `False` results in a `setVisible(True)`
2. This code currently only toggles visibility of `TyphosSignalPanel` descendents: https://github.com/pcdshub/typhos/blob/408c10892f25398b9e79f7cc9286c273ac918b6e/typhos/display.py#L1599

<img width="424" alt="image" src="https://user-images.githubusercontent.com/5139267/161608780-f7e7d396-9f75-478c-ac5c-a19ea4bcf614.png">

## Possible Solution
1. Keep track of user request for visibility change for each widget, and toggle based on that
2. Change the `findChildren` aggregation to better match widgets that should be hidden

## Steps to Reproduce (for bugs)

1. Load `at2l0` and click the top-level title. It should work to toggle everything below.
2. Try the title `at2l0_blade_02_state` and see it not work as expected

## Context
Long-standing bug that interferes with the functionality such that no one probably uses it

## Your Environment
pcds-5.3.0

PR #485, opened 2022-04-04
WIP/FIX: tweak toggle_display (not fully working)

klauer:
## Motivation and Context
Attempts to address #484 
... but fails to do so. Input/thought would be appreciated.

### DIFF
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -5,6 +5,7 @@ import inspect
 import logging
 import os
 import pathlib
+import weakref
 import webbrowser
 from typing import List, Optional, Union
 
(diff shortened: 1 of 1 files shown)


ZLLentz:
My first thought is that we should be toggling the visibility of a container widget instead of the individual visibilities of all the children. That would simply the logic a lot.

klauer:
I agree with the idea overall. Determining what is a container widget will probably require some further thought, unless I missed an obvious solution back when implementing this (which is entirely possible).

klauer:
Closing this as stale

PR #494, opened 2022-05-21
BUG: Fix Positioner clipping

ZLLentz:
## Description

This changes the sizing strategy of the positioner widget from what was essentially fixed-width sizing to minimum size and expanding. This fixes cases where the positioner widget looked scrunched up or otherwise clipping.

## Motivation and Context

It was reported in RIX that large numbers in the limits fields (in this case, 12 visible digits) would get clipped at the ends when rendered on the 4K high-dpi displays in the control room.

## How Has This Been Tested?

Interactively only- need to run on the RIX control room machines to fully confirm.

Just this PR

I don't have a high-dpi monitor at my desk, so to simulate and test this I would set the font side of the limits fields that were clipping on the control room display from 8pt to 50pt to guarantee clipping.

Here is what it looked like before:
before-yesclip

And after:
after-noclip

### DIFF
--- a/typhos/ui/widgets/positioner.ui
+++ b/typhos/ui/widgets/positioner.ui
@@ -60,8 +60,8 @@
      </property>
      <property name="baseSize">
       <size>
-       <width>400</width>
-       <height>250</height>
+       <width>100</width>
+       <height>100</height>
       </size>
      </property>
      <property name="frameShape">
--- a/typhos/ui/widgets/positioner.ui
+++ b/typhos/ui/widgets/positioner.ui
@@ -473,7 +473,7 @@ Screen</string>
            <string/>
           </property>
           <property name="text">
-           <string/>
+           <string>error_label</string>
           </property>
           <property name="wordWrap">
            <bool>true</bool>
--- a/typhos/ui/widgets/positioner.ui
+++ b/typhos/ui/widgets/positioner.ui
@@ -47,12 +47,18 @@
    <item alignment="Qt::AlignHCenter">
     <widget class="QFrame" name="motion_frame">
      <property name="sizePolicy">
-      <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+      <sizepolicy hsizetype="Minimum" vsizetype="Minimum">
        <horstretch>0</horstretch>
        <verstretch>0</verstretch>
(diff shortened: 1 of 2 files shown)


typhos/ui/widgets/positioner.ui
```diff
@@ -467,7 +473,7 @@ Screen</string>
            <string/>
           </property>
           <property name="text">
-           <string>error_label</string>
+           <string/>
```
ZLLentz, line 476: I have no idea what happened here. I'm tempted to revert this specific change manually. I'll spend a moment figuring out what it actually is.

ZLLentz: Reverted

typhos/ui/widgets/positioner.ui
```diff
@@ -47,12 +47,18 @@
...
-      <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+      <sizepolicy hsizetype="Minimum" vsizetype="Minimum">
        <horstretch>0</horstretch>
        <verstretch>0</verstretch>
       </sizepolicy>
      </property>
      <property name="maximumSize">
+      <size>
+       <width>16777215</width>
+       <height>16777215</height>
```
tangkong, line 58: Was there a reason for this number?  Asking out of curiosity is all.

tangkong (approved):
Looks good to me, works for my env and looks nice.

typhos/ui/widgets/positioner.ui
ZLLentz: This is the maximum number that designer lets you put in.

ZLLentz:
I might do one more small change and drop the minimums down a bit smaller and see what the effects are, in case they help with cases where the widget can be smaller than the number I picked.

ZLLentz:
> automated test screenshot artifacts for both low and high DPI displays

This is sort of possible to emulate for text by doing some math and rescaling everything. You can take a screenshot with the point sizes of your local machine with your local dpi, do some math and figure out what the pixel sizes would be at other dpis. I think it might be feasible.

ZLLentz:
Solving that problem though might be equivalent to finding a way for pydm to display all fonts the same size anyway

PR #496, opened 2022-05-24
BUG: remove word wrap from alarm field to fix rendering issue in rix

ZLLentz:
## Description

One small follow-up to the clipping fixes in #494 
"no alarm" is often rendering as "no", with alarm being clipped at the newline. Force "no alarm" to be one line by disabling word wrap on this QLabel.

### DIFF
--- a/typhos/ui/widgets/positioner.ui
+++ b/typhos/ui/widgets/positioner.ui
@@ -659,7 +659,7 @@ Screen</string>
              <set>Qt::AlignCenter</set>
             </property>
             <property name="wordWrap">
-             <bool>true</bool>
+             <bool>false</bool>
             </property>
            </widget>
           </item>


ZLLentz:
on second thought, I'm not sure about this as a solution- might close because it causes some sizing oddities
I need to understand why the "no alarm" text is clipping first, it wasn't obvious at first look and I didn't have a lot of time set aside to sit in RIX to figure out the specifics

klauer:
Just a guess -

Maybe size policies are being a problem here. The `alarm_label` is set to the default size policy of [preferred]():
> Preferred: The sizeHint() is best, but the widget can be shrunk and still be useful. The widget can be expanded, but there is no advantage to it being larger than sizeHint() (the default [QWidget](https://doc.qt.io/qt-5/qwidget.html) policy).

Should it not be minimum (or even fixed based on the contents)?

ZLLentz:
you're right, it cannot be shrunk and still be useful, so it should not be preferred

ZLLentz:
Some more thinking: having size differences between different text in the same box will continue to cause annoying problems as the gui shifts around when alarms are set and cleared. I think part of the solution needs to include normalizing the text used in this field so that the layout doesn't change when the text is changed.

ZLLentz:
This is no longer needed after #611

PR #498, opened 2022-05-26
FIX: show python documentation when no web help is available

klauer:
## Description
* Only show relevant help items when TYPHOS_HELP_URL is set
* Otherwise, show docstring-related information as it's still potentially useful

## Motivation and Context
Closes #452

## How Has This Been Tested?
Interactively

## Where Has This Been Documented?
Issue + PR text

## Screenshots (if appropriate):
image
image

PR #499, opened 2022-05-26
FIX: suite side popbar width

klauer:
## Description
Initial attempt at fixing the suite sidebar

## Motivation and Context
The device tree isn't really visible, presumably due to either typhos refactoring or pyqtgraph changes relating to Qt sizing/layouts.

I think this isn't the correct solution, but it'll work as a temporary solution as I investigate further.
This also doesn't affect the overlay not being sized properly (when hovering over the popbar but not pinning it). That remains broken as of the first commit.

### Master
image

### This PR
image

### DIFF
--- a/typhos/suite.py
+++ b/typhos/suite.py
@@ -198,6 +198,12 @@ class TyphosSuite(TyphosBase):
         self._bar = pcdsutils.qt.QPopBar(title='Suite', parent=self,
                                          widget=self._tree, pin=pin)
 
+        self._tree.setSizePolicy(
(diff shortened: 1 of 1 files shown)


ZLLentz:
Is the diff here doubled up with #498 ?

ZLLentz:
I think https://github.com/pcdshub/typhos/pull/499/commits/58ee2ad6123c41fd8d4c780db0d183e42e9dab28 is the functional commit which looks totally reasonable

klauer:
Oops, `git` strikes again

ZLLentz (approved):
I see no reason not to merge this

PR #505, opened 2022-06-01
FIX: profile classmethods as well

klauer:
## Description
Currently, `@classmethod` annotated class methods are not picked up in the profiling tools.
This fix includes them in the profiler output.

## Motivation and Context
I noticed the cached path tooling wasn't working as expected.

## How Has This Been Tested?
Locally
```
$ typhos --profile-modules typhos pydm at2l0_calc
```

This PR text

### DIFF
--- a/typhos/benchmark/utils.py
+++ b/typhos/benchmark/utils.py
@@ -2,13 +2,14 @@
 Helpful functions that don't belong in a more specific submodule.
 """
 import importlib
+import inspect
 import logging
 import os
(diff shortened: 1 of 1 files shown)


ZLLentz:
This looks fine, but:
- Why is this segfaulting the travis on 3.7/3.8? Can't possibly be related, right?
- Is it better to address this in `pcdsutils` and switch typhos to a pcdsutils import?

PR #491, opened 2022-05-13
REF: Designer widgets and data plugins using pydm's entrypoints

klauer:
## Description
* Removes custom designer widget plugin, instead relying on PyDM's own mechanism
* Activation scripts are removed
* Removes on-import data plugin instantiation. Now only applicable when pydm instantiates its data plugins.
* conda-forge recipe will need updating

Upstream PR: https://github.com/slaclab/pydm/pull/861

## Motivation and Context
Activation scripts are painful to write, maintain, and even use sometimes.
Let's remove an additional source of failure and rely on PyDM where it's sensible to.

Closes #504

## How Has This Been Tested?
Interactively

## Where Has This Been Documented?
This PR text

## Screenshots (if appropriate):
With the groups set as of 4cf5612:
image

PR #510, opened 2022-06-07
TST: fix ophyd signal value deprecation warning

klauer:
## Description
Removes deprecated and annoying `.value` access in test suite for ophyd

## Motivation and Context
Pulled out of #491 to isolate its issues

## How Has This Been Tested?
Test suite

This PR text


ZLLentz (approved):
This is the closest we'll ever get to a PR that can be merged without review

tangkong:
I'm surprised it was only the one instance of .value

PR #519, opened 2022-09-22
BUG: do not wait for lazy connections in SignalPanel

tangkong:
## Description
- Do not wait for lazy signals when creating a `SignalPanel`

## Motivation and Context
Found in working with lightpath that devices with area detectors would wait for every signal to time out when creating a detailed screen. https://github.com/pcdshub/lightpath/issues/161

The issue could be solved by enforcing no_lazy_wait_for_connection at device creation, but this catches devices regardless of how they were created

## How Has This Been Tested?
Works in lightpath interactively. 
Test suite passed locally, but severely slowed the typing of this pr text

This PR text.

### DIFF
--- a/typhos/panel.py
+++ b/typhos/panel.py
@@ -611,7 +611,8 @@ class SignalPanel(QtWidgets.QGridLayout):
 
         if self._should_show(kind, dotted_name, **self._filter_settings):
             try:
-                signal = getattr(device, dotted_name)
+                with ophyd.do_not_wait_for_lazy_connection(device):
+                    signal = getattr(device, dotted_name)
             except Exception as ex:
                 logger.warning('Failed to get signal %r from device %s: %s',
                                dotted_name, device.name, ex, exc_info=True)


ZLLentz:
> Test suite passed locally, but severely slowed the typing of this pr text

I did the same thing to myself last time I worked on `typhos`. The test suite pops up windows that briefly take the focus which is very disruptive to ongoing typing

ZLLentz (approved):
LGTM

tangkong:
I forgot about this little bug fix.  I'll merge this tomorrow if there's no objections

klauer (approved):
I'm not really sure who has the responsibility to set up the lazy-do-not-wait setting, to be honest. Is this the right spot? Should it be up to the caller to handle? Should it have been set at the ophyd level prior to any of this because it's a GUI application? 🤷‍♂️ 

Because I have no better answer and this fixes it for you, I'd say 👍 let's just move forward with this.

Edit: the other spot we do this is at the CLI level, just here: https://github.com/klauer/typhos/blob/14af093e703baccfe4e18c2eef5d394d9213c8fe/typhos/cli.py#L504 but it's insufficient

tangkong:
When I first ran into this I think I had a reason for putting it here, but time erodes all.

PR #520, opened 2022-10-19
FIX: stop causing unknown status error from user stop

ZLLentz:
## Description

Make sure we call stop with success=True in the positioner widget when the user clicks on the stop button.

## Motivation and Context

If we call stop with success=False, it marks the status as failed and gives us an unknown status error. This is a problem because we can confuse ourselves immediately. The screen should not report planned stops as unknown errors.

`success` is an expected keyword in the `bluesky` hardware interfaces: https://blueskyproject.io/bluesky/hardware.html#bluesky.protocols.Stoppable.stop

Here is the ophyd source path where the status gets marked as failed if we call with success=False (the default):
- https://github.com/bluesky/ophyd/blob/0e8acb3df3d17e0ab7aa2cc924831f4a0c580449/ophyd/positioner.py#L107
- https://github.com/bluesky/ophyd/blob/master/ophyd/positioner.py#L218-L220
- https://github.com/bluesky/ophyd/blob/0e8acb3df3d17e0ab7aa2cc924831f4a0c580449/ophyd/positioner.py#L209
- https://github.com/bluesky/ophyd/blob/0e8acb3df3d17e0ab7aa2cc924831f4a0c580449/ophyd/status.py#L374-L381

## How Has This Been Tested?

I'll find a way to test this

Not yet, I need to check to see if this warrants a docs addition.

### DIFF
--- a/typhos/positioner.py
+++ b/typhos/positioner.py
@@ -225,7 +225,7 @@ class TyphosPositionerWidget(utils.TyphosBase, widgets.TyphosDesignerMixin):
     def stop(self):
         """Stop device"""
         for device in self.devices:
-            device.stop()
+            device.stop(success=True)
 
     @QtCore.Slot()
     def clear_error(self):


ZLLentz:
Two tests fail due to a small oversight in pydm v1.17.0 that is fixed in master. I'm going to xfail those tests for now and move forward.

klauer (approved):
This looks good to me.

I restarted the Travis job as it failed unrelated to these changes

ZLLentz:
The travis keeps failing because of a bug in pydm v1.17.0, but maybe you saw a different issue

typhos/tests/conftest.py
```diff
@@ -283,3 +283,10 @@ def reset_signal_plugin():
 @pytest.fixture(scope='function', autouse=True)
 def show_test_name(request):
     logger.debug(f'Running test named {request.node.name}')
+
+
+# Remove this later please
+pydm_version_xfail = pytest.mark.xfail(
+    condition=pydm.__version__ == '1.17.0',
+    reason='PyDM bug in v1.17.0',
+)
```
ZLLentz, line 292: I think I like this pattern a bit more than the pain of changing the recipes repeatedly... even though it's a bit haphazard

klauer: I like it 👍

tangkong (approved):
LGTM too, pending tests passing / getting skipped

PR #546, opened 2023-02-15
Disambiguate duplicate template filenames

### LINKED ISSUE #524: Disambiguate duplicate template filenames
## Expected Behavior

If multiple files are found for the same device class, we should expand them to display the full path to the user.

## Current Behavior

https://github.com/pcdshub/engineering_tools/pull/139#pullrequestreview-1166010224

nrwslac:
## Description

- add util `find_duplicate_filenames_in_paths`
- check for duplicate filenames in `generate_context_menu` and `_generate_template_menu`
- set duplicate file names to full path

## Motivation and Context

When two templates with the same name are found in two different directories, typhos displays identical names. This makes it impossible to tell which template is being selected.

- closes #524 

## How Has This Been Tested?

I tested this using happi entry `mr2k4_kbo`. I put 3 screens in the `typhos` core ui directory using the keywords embedded, detailed, and engineering. I put identical named files into my local `pcdsdevices` ui directory. I added both my local libraries to the `$PYTHONPATH` and activated the current `pcds_conda`: pcds-5.6.0. I then ran `typhos mr2k4_kbo` and checked each of the four menu buttons by right clicking. Each right click showed full path for identical files, and filenames otherwise. I checked embedded, engineering, detailed, and the template button.

- added/updated doc strings in created/modified functions

typhos-context

### DIFF
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -1118,14 +1118,14 @@ class TyphosDeviceDisplay(utils.TyphosBase, widgets.TyphosDesignerMixin,
             duplicates = utils.find_duplicate_filenames_in_paths(filenames)
 
             for filename in filenames:
-                current_filename = os.path.split(filename)[-1]
+                current_filename = os.path.basename(filename)
 
                 def switch_template(*, filename=filename):
                     self.force_template = filename
                 if current_filename in duplicates:
                     action = menu.addAction(str(filename))
                 else:
-                    action = menu.addAction(os.path.split(filename)[-1])
+                    action = menu.addAction(current_filename)
                 action.triggered.connect(switch_template)
 
         refresh_action = base_menu.addAction("Refresh Templates")
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -419,11 +419,16 @@ class TyphosDisplaySwitcherButton(TyphosToolButton):
             return
 
         menu = QtWidgets.QMenu(parent=self)
+
+        duplicates = utils.find_duplicate_filenames_in_paths(self.templates)
+
         for template in self.templates:
             def selected(*, template=template):
                 self.template_selected.emit(template)
-
-            action = menu.addAction(template.name)
+            if template.name in duplicates:
+                action = menu.addAction(str(template))
+            else:
+                action = menu.addAction(template.name)
             action.triggered.connect(selected)
 
         return menu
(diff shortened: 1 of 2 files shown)


typhos/display.py
```diff
@@ -1110,11 +1115,17 @@ def _generate_template_menu(self, base_menu):
...
 
+            duplicates = utils.find_duplicate_filenames_in_paths(filenames)
+
             for filename in filenames:
+                current_filename = os.path.split(filename)[-1]
+
                 def switch_template(*, filename=filename):
                     self.force_template = filename
-
-                action = menu.addAction(os.path.split(filename)[-1])
```
ZLLentz, line 1117: The old code could have used `os.path.basename(filename)` instead of picking the last (second) element from the path split, and the new code can do this too since you don't need the directory name

ZLLentz (commented):
I think this is a good approach

typhos/display.py
nrwslac: good one, added.

tangkong (commented):
I think this looks good on the whole, though we should talk about your screenshot technique 😆 

The CI test failures are unrelated to this PR, and have been observed elsewhere (and will hopefully be addressed elsewhere...).

The duplicate filename util function doesn't group absolute paths and relative paths, but in this case it's not a problem I believe

nrwslac:
> I think this looks good on the whole, though we should talk about your screenshot technique 😆

Uploaded a more manageable screen shot 📸

typhos/display.py
```diff
@@ -414,16 +414,22 @@ def _clicked(self):
         self.template_selected.emit(template)
 
     def generate_context_menu(self):
-        """Context menu request."""
+        """Context menu request.
+            Duplicate filenames will show full path. """
```
ZLLentz, line 418: The multi-line docstring here isn't indented correctly, but it could have stayed as a single-liner
```suggestion
        """Context menu request. Duplicate filenames will show full path. """
```

typhos/display.py
```diff
@@ -1104,17 +1110,26 @@ def _move_display_to_layout(self, widget):
         self._scroll_area.setVisible(scrollable)
 
     def _generate_template_menu(self, base_menu):
-        """Generate the template switcher menu, adding it to ``base_menu``."""
+        """
+            Generate the template switcher menu, adding it to ``base_menu``.
+            Duplicate filenames will show full path.
+        """
```
ZLLentz, line 1116: Multi-line docstring conventions have the indentation like this
```suggestion
        """
        Generate the template switcher menu, adding it to ``base_menu``.
        Duplicate filenames will show full path.
        """
```

ZLLentz (commented):
Docstring nitpicks
Otherwise I think this is good to go

Only thing lacking is no tests added, some parts of this are kind of annoying to write unit tests for though

klauer (changes requested):
Nicely done - a great first contribution to typhos, Nick! You found the right spot and fixed it up in a clean way.

A few remaining formatting things and this should be good-to-go.

typhos/utils.py
```diff
@@ -270,6 +269,27 @@ def random_color():
                   random.randint(0, 255))
 
 
+def find_duplicate_filenames_in_paths(paths):
+    """
+    Returns a list of duplicated filenames from a list of full path filenames.
```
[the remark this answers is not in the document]
nrwslac: I naively thought pre-commit would format my doc strings for me :rofl:. I'll take a look at the docs and work on fixing this up, thank you.

nrwslac:
> Only thing lacking is no tests added, some parts of this are kind of annoying to write unit tests for though

Probably worth taking some time to look at how the testing works for this repo.

klauer:
I forgot this was a thing and we never got around to fixing and merging, sorry @nrwslac!

This PR is superseded by #572

klauer:
#572 was merged; closing

PR #548, opened 2023-02-22
FIX: missing and extra device fills

ZLLentz:
## Description

- In templates, alarm shapes auto-fill the template's device now
- When loading a template, avoid overwriting user-set happi channels and other similar devices

## Motivation and Context

https://jira.slac.stanford.edu/browse/LCLSPC-653

## How Has This Been Tested?

I will add a test prior to asking for reviewers

Comments, and later, release notes


ZLLentz:
This PR will need to expand in scope a bit: I can't test it properly without adding some small features that I want to add anyway. The issue is that it's not currently possible to apply subdevices to widgets if they are not sourced from happi.

ZLLentz:
This PR is no longer relevant, I will close it

PR #550, opened 2023-02-23
FIX: include pydm and user pydm stylesheets

### LINKED ISSUE #540: Typhos ignores PyDM stylesheet settings
## Expected Behavior

Consistency with user's custom PyDM style and built-in PyDM style as appropriate

## Current Behavior

Typhos fully overwrites all stylesheet settings

## Possible Solution

Use the pydm stylesheet utilities from https://github.com/slaclab/pydm/blob/229464274c5b052ca8efb3d3053f6099bca10bfa/pydm/utilities/stylesheet.py

ZLLentz:
## Description

- Include the normal stylesheets that load when `pydm` loads.
- Re-implement the stylesheet loading and composition rather than using `pydm`'s utilities to enforce a specific order.
- Include a way to add a stylesheet in CLI instead of the previous version which only had a way to override all stylesheets.

## Motivation and Context

closes #540

## How Has This Been Tested?

Added a test, tested interactively with the case I cared about

Docstrings, comments, and sphinx docs updating.
Later: release notes

### DIFF
--- a/typhos/cli.py
+++ b/typhos/cli.py
@@ -20,7 +20,7 @@ from typhos.benchmark.cases import run_benchmarks
 from typhos.benchmark.profile import profiler_context
 from typhos.display import DisplayTypes, ScrollOptions
 from typhos.suite import TyphosSuite
-from typhos.utils import nullcontext
+from typhos.utils import compose_stylesheets, nullcontext
 
 logger = logging.getLogger(__name__)
 
@@ -190,15 +190,12 @@ def typhos_cli_setup(args):
     coloredlogs.install(level=level, logger=shown_logger, fmt=log_fmt)
     logger.debug("Set logging level of %r to %r", shown_logger.name, level)
 
-    # Deal with stylesheet
-    qapp = get_qapp()
-
     logger.debug("Applying stylesheet ...")
-    typhos.use_stylesheet(dark=args.dark)
-    if args.stylesheet:
-        logger.info("Loading QSS file %r ...", args.stylesheet)
-        with open(args.stylesheet) as handle:
-            qapp.setStyleSheet(handle.read())
+    compose_stylesheets(
+        dark=args.dark,
+        path=args.stylesheet,
+        widget=get_qapp(),
+    )
 
 
 def _create_happi_client(cfg):
--- a/typhos/cli.py
+++ b/typhos/cli.py
@@ -20,7 +20,8 @@ from typhos.benchmark.cases import run_benchmarks
 from typhos.benchmark.profile import profiler_context
 from typhos.display import DisplayTypes, ScrollOptions
 from typhos.suite import TyphosSuite
-from typhos.utils import compose_stylesheets, nullcontext
+from typhos.utils import (apply_standard_stylesheets, compose_stylesheets,
+                          nullcontext)
 
 logger = logging.getLogger(__name__)
 
@@ -131,10 +132,12 @@ parser.add_argument(
 )
 parser.add_argument(
     "--stylesheet-override", "--stylesheet",
+    type="append",
     help="Override all built-in stylesheets, using this stylesheet instead.",
 )
 parser.add_argument(
     "--stylesheet-add",
+    type="append",
     help=(
         "Include an additional stylesheet in the loading process. "
         "This stylesheet will take priority over all built-in stylesheets, "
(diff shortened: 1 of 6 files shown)


typhos/cli.py
```diff
@@ -129,7 +129,18 @@
...
+parser.add_argument(
+    "--stylesheet-override", "--stylesheet",
+    help="Override all built-in stylesheets, using this stylesheet instead.",
+)
+parser.add_argument(
+    "--stylesheet-add",
+    help=(
+        "Include an additional stylesheet in the loading process. "
+        "This stylesheet will take priority over all built-in stylesheets, "
+        "but not over a template or widget's styleSheet property."
```
tangkong, line 141: Not a bug, but this argument description read to me as allowing the user to apply multiple stylesheets by invoking this argument multiple times.  If you do this, only the last `--stylesheet-add STYLESHEET` will be applied, and maybe more confusingly there's no error thrown.  

* The simple option is to just add a little bit to the help documentation notifying users that the option should only be used once. 
* A more involved path is to support multiple stylesheets added using this option, using `type='append'`, and modifying `compose_stylesheets` to take lists of stylesheets.

I like the second option, but I'd also be surprised if people were trying to set all their stylesheets through the cli when the environment variable settings exist.

klauer: I'd lean toward the second as well, since we're touching on this code anyway.

It seems not a big ask to switch `path` to `paths` in the `compose_stylesheets` function.

typhos/utils.py
```diff
@@ -228,38 +234,104 @@ def clean_name(device, strip_parent=True):
...
-        # Load the stylesheet from the file
         with open(style_path) as handle:
-            style = handle.read()
-    if widget is None:
-        widget = QtWidgets.QApplication.instance()
-    # We can set Fusion style if it is an application
+            return handle.read()
+
+
+def compose_stylesheets(
```
klauer, line 277: Code reusability-wise, I'd have expected that `compose_stylesheets` would merely have combined the source code so that you could do with it as you please.

A more apt name might be "compose stylesheets and apply to maybe a widget or the entire application" which makes it sound like it's doing a bit too much.

klauer (approved):
The approach looks good and I think this is sufficient to fix the issue.

Though there are a couple smelly bits, I think this is probably good to go.

I'll need to take a look at reusing this code in downstream GUIs as well, replacing my own attempts.

typhos/cli.py
ZLLentz: The truth is that I wouldn't have included this arg at all if I was redoing this from nothing, but since there was an existing "override the stylesheet with one path" I extended it slightly.
It's not a big ask to support multiple stylesheets from the arg here as you both describe :+1:

typhos/utils.py
ZLLentz: I agree.

It turns out that the only thing you need to do to combine two stylesheets is to append them together, but having a function to do that and to abstract away the priority order could be useful.

ZLLentz: I'll rename and refactor out the order/priority bits to a smaller function with this original name.

ZLLentz: For more context: there are related utilities in https://github.com/slaclab/pydm/blob/master/pydm/utilities/stylesheet.py but they are a bit too specific to be used as-is here

ZLLentz:
> a couple smelly bits

Are there any other smelly bits that you frowned at and passed over? Or is it mainly the code reusibility of the stylesheet helper functions?

klauer:
> Are there any other smelly bits that you frowned at and passed over? Or is it mainly the code reusibility of the stylesheet helper functions?

Just the above things we had commented on - both very minor in the grand scheme of things.

ZLLentz:
Future us will be happier with less smelly repos, so since these are easy and in scope I'll add them on

ZLLentz:
Unless I broke it further with that commit, what remains is adjusting that test to call the new function name and adding a test for the refactored `compose_stylesheets` to check priority order.

ZLLentz:
I've done the minor refactor, added a new test, and cleaned up a bit

typhos/utils.py
```diff
@@ -308,16 +308,20 @@ def compose_stylesheets(stylesheets: Iterable[str | pathlib.Path]) -> str:
...
-        If any error is encountered while reading a file.
+        If any error is encountered while reading a file
+    TypeError
+        If the input is not a valid type
     """
     style_parts = []
     for sheet in stylesheets:
         path = pathlib.Path(sheet)
-        if path.suffix == "qss":
+        if isinstance(sheet, pathlib.Path) or path.suffix == ".qss":
```
tangkong, line 319: oh  my that's a pesky `.`

ZLLentz:
I'll merge some time tomorrow unless either of you want to re-review to improve the refactor

klauer:
I think you addressed everything 👍  LGTM

PR #557, opened 2023-07-11
ENH: add notes field in TyphosDisplayTitle

tangkong:
## Description
One proposal for how one might add descriptions to devices.

Searches in the following places for a device notes yaml:
1. `<user_data_path>/device_notes.yaml`
2. `PCDS_DEVICE_NOTES` environment variable

The above list is also the priority ranking, where higher priority sources shadow/overwrite lower priority information.  (aka if both of the above exist, information from `DEVICE_NOTES` will take precedence).  In general more local information will shadow more global information

If nothing exists, we default to the user_data_path, where we also write new note information 

All of this is up for debate and subject to change.

## Motivation and Context
https://jira.slac.stanford.edu/browse/ECS-1039 , but also dovetailing off of https://github.com/pcdshub/lucid/pull/124

## How Has This Been Tested?
interactively, opening typhos and lucid

this PR.

image

image

### DIFF
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -21,6 +21,7 @@ from . import cache
 from . import panel as typhos_panel
 from . import utils, web, widgets
 from .jira import TyphosJiraIssueWidget
+from .notes import TyphosNotesEdit
 from .plugins.core import register_signal
 
 logger = logging.getLogger(__name__)
@@ -805,10 +806,13 @@ class TyphosDisplayTitle(QtWidgets.QFrame, widgets.TyphosDesignerMixin):
         self.underline.setFrameShadow(self.underline.Plain)
         self.underline.setLineWidth(10)
 
+        self.notes_edit = TyphosNotesEdit()
+
         self.grid_layout = QtWidgets.QGridLayout()
         self.grid_layout.addWidget(self.label, 0, 0)
-        self.grid_layout.addWidget(self.switcher, 0, 1, Qt.AlignRight)
-        self.grid_layout.addWidget(self.underline, 1, 0, 1, 2)
+        self.grid_layout.addWidget(self.switcher, 0, 2, Qt.AlignRight)
+        self.grid_layout.addWidget(self.notes_edit, 0, 1, Qt.AlignLeft)
+        self.grid_layout.addWidget(self.underline, 1, 0, 1, 3)
 
         self.help = TyphosHelpFrame()
         if utils.HELP_WEB_ENABLED:
@@ -878,6 +882,9 @@ class TyphosDisplayTitle(QtWidgets.QFrame, widgets.TyphosDesignerMixin):
         """Typhos hook for setting the associated device."""
         if not self.label.text():
             self.label.setText(device.name)
+
+        if not self.notes_edit.text():
+            self.notes_edit.setup_data(device.name)
 
         if self.help is not None:
             self.help.add_device(device)
--- a/typhos/notes.py
+++ b/typhos/notes.py
@@ -120,7 +120,7 @@ class TyphosNotesEdit(QtWidgets.QLineEdit):
 
     def update_tooltip(self) -> None:
         if self.data:
-            self.setToolTip(f"({self.data['timestamp']}):\nu{self.data['note']}")
+            self.setToolTip(f"({self.data['timestamp']}):\n{self.data['note']}")
 
     def setup_data(self, device_name: str) -> None:
         """
--- a/typhos/notes.py
+++ b/typhos/notes.py
@@ -1,0 +1,153 @@
+import logging
+import os
+from datetime import datetime
+from enum import IntEnum
+from pathlib import Path
+from typing import Dict, Tuple
+
+import platformdirs
+import yaml
+from qtpy import QtWidgets
+
+from typhos import utils
+
+logger = logging.getLogger(__name__)
+
+
+class NotesSource(IntEnum):
+    USER = 0
+    ENV = 1
+    HAPPI = 2
+
+
+def get_notes_data(device_name: str) -> Tuple[NotesSource, Dict[str, str]]:
+    """
+    get the notes data for a given device
+    attempt to get the info from the following locations
+    in order of priority (higher priority will shadow)
+    1: device_notes in the user directory
+    2: the path in DEVICE_NOTES environment variable
+    3: TODO: happi ... eventually
+
+    Parameters
+    ----------
+    device_name : str
+        The name of the device to retrieve notes for.
+
+    Returns
+    -------
+    Tuple[NotesSource, dict]
+        The source of the device notes
+        a dictionary containing the device note information
+    """
+    data = {'note': '', 'timestamp': ''}
+    source = NotesSource.USER
+    # try user directory
+    user_data_path = platformdirs.user_data_path() / 'device_notes.yaml'
+    if user_data_path.exists():
+        with open(user_data_path) as f:
+            device_notes = yaml.full_load(f)
+
+        if device_name in device_notes:
+            source = NotesSource.USER
+            data = device_notes.get(device_name)
+
+    env_notes_path = os.environ.get('DEVICE_NOTES')
+    if env_notes_path and Path(env_notes_path).is_file():
+        with open(Path(env_notes_path)) as f:
+            device_notes = yaml.full_load(f)
+
+        if device_name in device_notes:
+            source = NotesSource.ENV
+            data = device_notes.get(device_name)
+
+    return source, data
+
+
+def insert_into_data(path: Path, device_name: str, data: dict[str, str]) -> None:
+    with open(path, 'r') as f:
+        device_notes = yaml.full_load(f)
+
+    device_notes[device_name] = data
+
+    with open(path, 'w') as f:
+        yaml.dump(device_notes, f)
+
+
+def write_notes_data(
+    source: NotesSource,
+    device_name: str,
+    data: dict[str, str]
+) -> None:
+    """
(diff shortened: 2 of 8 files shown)


typhos/notes.py
```diff
@@ -0,0 +1,153 @@
...
+    user_data_path = platformdirs.user_data_path() / 'device_notes.yaml'
+    if user_data_path.exists():
+        with open(user_data_path) as f:
+            device_notes = yaml.full_load(f)
+
+        if device_name in device_notes:
+            source = NotesSource.USER
+            data = device_notes.get(device_name)
+
+    env_notes_path = os.environ.get('DEVICE_NOTES')
```
klauer, line 55: Can we tie this filename to typhos a bit more?
Can we share this information with lucid and use the same file?

Since lucid depends on typhos, it'd be OK if lucid supports typhos's format (and not the other way around) I think.

typhos/notes.py
```diff
@@ -0,0 +1,153 @@
...
+        Parameters
+        ----------
+        device_name : str
+            The device name.  Can also be a component. e.g. device_component_name
+        """
+        if self.device_name:
+            return
+
+        self.device_name = device_name
+        self.notes_source, self.data = get_notes_data(device_name)
```
klauer, line 142: It'd be good if syntax errors / failure to load the yaml file didn't cause typhos to crash / fail to load a display

typhos/notes.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    if source == NotesSource.USER:
+        user_data_path = platformdirs.user_data_path() / 'device_notes.yaml'
+        insert_into_data(user_data_path, device_name, data)
+    elif source == NotesSource.ENV:
+        env_data_path = Path(os.environ.get('DEVICE_NOTES'))
+        insert_into_data(env_data_path, device_name, data)
+
+
+class TyphosNotesEdit(QtWidgets.QLineEdit):
```
klauer, line 102: Is the intention single-line notes and not multi-line?

typhos/notes.py
tangkong: Perhaps in the interest of using this for multiple apps we keep this more general?  (since we talked about using this apps like hutch-python, etc)

klauer: An environment variable without a shared prefix still doesn't sound great to me (even if we don't make it `TYPHOS_` something). It leaves me not knowing what to grep for, or what goes along with what when I type `env` to look at everything we define in our environments (which is a lot).

I know this sounds nitpicky on the surface, but changing it going forward can be also be painful/confusing. 
... So, can we do better than `DEVICE_NOTES`?

tangkong: Fair point.  But names are hard.  We don't currently have a prefix for variables shared between our libraries, which is what I think we want here.  Thoughts on:
* `PCDS_DEVICE_NOTES`: risks making alex sad about dead-naming our group, but matches our pcdshub org
* `ECS_DEVICE_NOTES`: more "modern", but it feels like I'm claiming too much... maybe that's just a me problem
* `LCLS_DEVICE_NOTES`: definitely too general of an environment variable

ZLLentz: My initial thought is that environment variables are easy enough to tweak that something like `TYPHOS_DEVICE_NOTES` would be totally OK. If this migrates to happi and becomes `HAPPI_DEVICE_NOTES` or if it goes away entirely and gets consumed by happi it'll be ok, and if we need to push it up to something like `pcdsutils` it can be renamed easily too. We can even continue to support `TYPHOS_DEVICE_NOTES` as a backup in whatever capacity we move it to.

Note that typhos doesn't necessarily imply pcds/ecs as well.

tangkong: As long as we decide on something before people start setting env vars in their various .bashrc's and don't have to go through correcting them, I think I agree.  Changing things in our control is easy enough, it's the general public I'm more worried about

tangkong: I have chosen `PCDS_DEVICE_NOTES`.  Hopefully future me will look back on this moment and not be too annoyed

typhos/notes.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+    try:
+        with open(path) as f:
+            device_notes = yaml.full_load(f)
+    except Exception as ex:
+        logger.warning(f'failed to load device notes: {ex}')
+        return
+
+    if device_name in device_notes:
+        return device_notes.get(device_name, None)
```
ZLLentz, line 49: The if statement is redundant with the `.get` here
```suggestion
    return device_notes.get(device_name, None)
```

typhos/notes.py
```diff
@@ -0,0 +1,189 @@
...
+            device_notes = yaml.full_load(f)
+    except Exception as ex:
+        logger.warning(f'unable to open existing device info: {ex}')
+        return
+
+    device_notes[device_name] = data
+
+    try:
+        with open(path, 'w') as f:
+            yaml.dump(device_notes, f)
```
ZLLentz, line 106: This should use the "make a second file and copy over the first" method to make sure we do an atomic write

typhos/tests/test_notes.py
```diff
@@ -0,0 +1,99 @@
...
+from pytestqt.qtbot import QtBot
+
+from typhos.notes import NOTES_VAR, TyphosNotesEdit
+
+from .conftest import MODULE_PATH
+
+
+@pytest.fixture(scope='function')
+def user_notes_path(monkeypatch, tmp_path: Path):
+    # copy user device_notes.yaml to a temp file
```
ZLLentz, line 17: This feels pedantic but it actually changed my understanding of the comment
```suggestion
    # copy user_device_notes.yaml to a temp file
```

typhos/utils.py
```diff
@@ -1651,3 +1651,56 @@ def raise_window(widget):
     window.raise_()
     window.activateWindow()
     window.setFocus()
+
+
+class FrameOnEditFilter(QtCore.QObject):
```
ZLLentz, line 1656: What's the breakpoint on moving something like this to a utility module vs vendoring it everywhere it is needed? It's pretty short and might want to be customized a bit so I'm not sure

ZLLentz (commented):
Added my thoughts from reading the code
I need to give this a spin myself

typhos/notes.py
tangkong: ah right I forgot about that

typhos/utils.py
tangkong: I did have to tweak it a bit actually, though maybe we should be putting things into `pcdswidgets` more often?  

(sidebar, using the `finishedEditing` signal requires the FocusIn/FocusOut events to pass through, which mucks around with some of the formatting)

typhos/notes.py
ZLLentz, line 1: This passes the feels/vibes test for sure, it felt good to use.

typhos/notes.py
```diff
@@ -0,0 +1,189 @@
...
+    return source, data
+
+
+def insert_into_yaml(path: Path, device_name: str, data: dict[str, str]) -> None:
+    try:
+        with open(path, 'r') as f:
+            device_notes = yaml.full_load(f)
+    except Exception as ex:
+        logger.warning(f'unable to open existing device info: {ex}')
+        return
```
ZLLentz, line 99: This requires the user to create their own yaml file, else you only get this exception when you try to edit the descriptions:
```
[2023-07-12 16:43:31] - WARNING - unable to open existing device info: [Errno 2] No such file or directory: '/cds/home/z/zlentz/.local/share/device_notes.yaml'
```
I think there should be situations where we write a starting file for the user, maybe when no configurations are found at all?

typhos/notes.py
```diff
@@ -0,0 +1,189 @@
...
+
+        # to be initialized later
+        self.device_name: str = None
+        self.notes_source: Optional[NotesSource] = None
+        self.data = {'note': '', 'timestamp': ''}
+
+    def update_tooltip(self) -> None:
+        if self.notes_source is not None:
+            self.setToolTip(f"({self.data['timestamp']}, {self.notes_source.name}):\n"
+                            f"{self.data['note']}")
```
ZLLentz, line 158: I think we can do slightly better for a default tooltip when no description has been written:
image

Maybe just some simple plaintext "Click here to edit the description" if the note is empty?

typhos/display.py
```diff
@@ -879,6 +883,9 @@ def add_device(self, device):
         if not self.label.text():
             self.label.setText(device.name)
 
+        if not self.notes_edit.text():
+            self.notes_edit.setup_data(device.name)
```
ZLLentz, line 887: Any thoughts on running this more often/periodically/when the file updates? I think this falls out of sync fairly quickly.

typhos/notes.py
tangkong: Ah yes, I had that set up before but the return snuck in.  Removing it should be enough here to write a new file if one doesn't exist

typhos/display.py
tangkong: You're thinking for long-running screens?  Or if people edit the file manually and want the screen to update?  I'm not too big a fan of watching files, but I suppose we could run it periodically.  

I imagined a world where people only edit this via GUI apps, but I guess I can't assume that's reality

ZLLentz: Imagine I open the screen on tmo-control, then you open it on tmo-daq, then James opens it on tmo-console. The changes can't propagate between these copies of the screen without a restart.

klauer (approved):
LGTM - notes will be very useful 👍

ZLLentz (approved):

PR #563, opened 2023-08-05
ENH: layout changes for compact complex screens + row positioner widget

klauer:
## Description

* Add 'single line positioner' variant
* Tweaked the logic for "should we show a 'detailed tree' at each device layer", now taking into account whether or not the tree is nested (i.e., is it a top-level device?)
* Tweaked some other layout things while attempting to squeeze in the space a bit more
* Other things probably too
* (TODO) Devices snuck their way in here but those need to be removed. *Solely for testing purposes.*
* "Notes" feature needs to be re-added for positioners

Tweaks to the layout of the row positioner widget by @ZLLentz :

*   Shorter text in positioner row alarm label
*   Status container always visible
*   Dynamic showing/hiding of status info
*   Infinite arrangement/sizing/size policy nitpicking
*   Drop the concept of the whole widget getting a colored border during/after moves
*   No alarm borders inside the widget
*   Add some more accurate placeholder templates for twincat states things
*   Remove typhos stylesheet for the error pydm label widget

TODO
- [x] Zach will modify layout
- [ ] Later: decide as to if there should be a new row category for templates? Aspect ratio buckets for templates?
- [ ] Get feedback from ECS, then scientists, ...
- [x] Remove device UI files from this repo and move them to pcdsdevices prior to merging

## Motivation and Context
Request was "always show the embedded screen!" (https://github.com/pcdshub/typhos/issues/532)
That unfortunately does not render well if you take it literally. I'm of the opinion that it shouldn't be taken literally and we should explore options like this one.

Oh, turns out there was an actual issue for this: closes #543 
I must have read it and forgotten about it. The single row display for motors was what we used by default back at NSLS-II with CSS, so this was something I always wanted to see in typhos and only recently found the inspiration for.

## How Has This Been Tested?
Minimally, with at2l0 and multi motor example

```python
import ophyd
from ophyd import Component, Device, EpicsMotor, EpicsSignal
from qtpy import QtCore, QtWidgets

from typhos.suite import TyphosSuite

app = QtWidgets.QApplication([])

class MyMotors(ophyd.Device):
    m1 = Component(EpicsMotor, "AT2L0:XTES:MMS:01")
    m2 = Component(EpicsMotor, "AT2L0:XTES:MMS:02")
    m3 = Component(EpicsMotor, "AT2L0:XTES:MMS:03")
    m4 = Component(EpicsMotor, "AT2L0:XTES:MMS:04")
    m5 = Component(EpicsMotor, "AT2L0:XTES:MMS:05")
[...]

PR #566, opened 2023-08-09
ENH: add `--screenshot` to the typhos CLI

klauer:
## Description
Add `typhos --screenshot filename_pattern` to take screenshots of typhos displays prior to exiting early (in combination with `--exit-after`).

## Motivation and Context
Larger changes to layouts coming in #563 
We should check screenshots from the current master and compare them with #563 to ensure that the latter is strictly an improvement

## How Has This Been Tested?
Test suite and interactively

This PR text and pre-release notes

## Screenshots

```
$ typhos --fake-device ophyd.EpicsMotor[] --screenshot "{device.name}.png" --exit-after 1
[2023-08-09 09:40:41] - WARNING - Exiting typhos early due to --exit-after=1.0 CLI argument.
[2023-08-09 09:40:41] - INFO - Saving screenshot of 'Typhos Suite - EpicsMotor': 'EpicsMotor' to 'EpicsMotor.png'
```

EpicsMotor

### DIFF
--- a/typhos/cli.py
+++ b/typhos/cli.py
@@ -1,4 +1,6 @@
 """This module defines the ``typhos`` command line utility"""
+from __future__ import annotations
+
 import argparse
 import ast
 import inspect
@@ -6,6 +8,8 @@ import logging
 import re
 import signal
 import sys
+import types
+import typing
 from typing import Optional
 
 import coloredlogs
@@ -24,6 +28,31 @@ from typhos.utils import (apply_standard_stylesheets, compose_stylesheets,
                           nullcontext)
 
 logger = logging.getLogger(__name__)
+
+
+class TyphosArguments(types.SimpleNamespace):
+    """Type hints for ``typhos`` CLI entrypoint arguments."""
+
+    devices: list[str]
+    layout: str
+    cols: int
+    display_type: str
+    scrollable: str
+    size: Optional[str]
+    hide_displays: bool
+    happi_cfg: Optional[str]
+    fake_device: bool
+    version: bool
+    verbose: bool
+    dark: bool
+    stylesheet_override: Optional[list[str]]
+    stylesheet_add: Optional[list[str]]
+    profile_modules: Optional[list[str]]
+    profile_output: Optional[str]
+    benchmark: Optional[list[str]]
+    exit_after: Optional[float]
+    screenshot_filename: Optional[str]
+
 
 # Argument Parser Setup
 parser = argparse.ArgumentParser(
@@ -179,6 +208,14 @@ parser.add_argument(
     help=(
         "(For profiling purposes) Exit typhos after the provided number of "
         "seconds"
+    ),
+)
+parser.add_argument(
+    '--screenshot',
+    dest="screenshot_filename",
+    help=(
+        "Save a screenshot of the generated display(s) prior to exiting to "
+        "this filename"
     ),
 )
 
@@ -484,7 +521,8 @@ def typhos_run(
     initial_size: Optional[str] = None,
     show_displays: bool = True,
     exit_after: Optional[float] = None,
-) -> QtWidgets.QMainWindow:
+    screenshot_filename: Optional[str] = None,
+) -> Optional[QtWidgets.QMainWindow]:
     """
     Run the central typhos part of typhos.
 
@@ -516,6 +554,8 @@ def typhos_run(
     show_displays : bool, optional
         If True (default), open all the included device displays.
         If False, do not open any of the displays.
+    screenshot_filename : str, optional
+        Save a screenshot to this file prior to exiting early.
 
     Returns
     -------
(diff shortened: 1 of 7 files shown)


typhos/suite.py
```diff
@@ -767,6 +767,63 @@ def save(self):
...
+            suite_title = self.windowTitle()
+            widget_title = display.windowTitle()
+            if image is None:
+                logger.warning(
+                    "Failed to take screenshot of device: %s in %s",
+                    device.name, suite_title,
+                )
+                continue
+
+            filename = filename_format.format(
```
klauer, line 813: This means that the CLI `--screenshot` can be formatted with the following variables:
`suite_title`, `widget_title`, `device`, `name` (or any attribute thereof)

This should be better communicated in the CLI description

typhos/suite.py
```diff
@@ -767,6 +767,63 @@ def save(self):
...
+
+        filenames = {}
+        for device in self.devices:
+            display = self.get_subdisplay(device)
+
+            if hasattr(display, "to_image"):
+                image = display.to_image()
+            else:
+                # This is a fallback for if/when we don't have a TyphosDisplay
+                image = utils.take_widget_screenshot(display)
```
tangkong, line 802: Do you know how often this happens?  I was under the impression most everything was a `TyphosDeviceDisplay`

tangkong (approved):
I think the results here speak for themselves.  A stray thought from me and not much else

typhos/suite.py
klauer: It shouldn't _ever_ happen in the current codebase. The truth is the following:
1. I added `utils.take_widget_screenshot` first, added tests and had it all working
2. Then I shifted to "it'd be better to save each device display" 
3. Then I recalled that I had already made `to_image` on `TyphosDeviceDisplay` (a few years ago perhaps?)
4. Then I needed to justify the work from (1) so this `else` clause stayed in

typhos/tests/test_suite.py
```diff
@@ -123,35 +123,51 @@ def test_device_parameter_tree(qtbot, motor, device):
...
     handle = io.StringIO()
     save_suite(suite, handle)
     handle.seek(0)
     devices = [device.name for device in suite.devices]
     assert str(devices) in handle.read()
 
 
-def test_suite_save(suite, monkeypatch):
+def test_suite_save_screenshot(suite: TyphosSuite, device: MockDevice):
+    with tempfile.NamedTemporaryFile(mode="wb") as fp:
```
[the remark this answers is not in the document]
klauer: pytest fixtures generally feel like the right way of doing things in the test suite - I'm just more familiar with this standard lib one. I'll have to peek at the `tmp_path` API a bit out of curiosity...

typhos/cli.py
```diff
@@ -25,6 +29,31 @@
 
 logger = logging.getLogger(__name__)
 
+
+class TyphosArguments(types.SimpleNamespace):
```
ZLLentz, line 33: I'm a big fan of using placeholder classes like this for arg type handling

typhos/cli.py
```diff
@@ -534,34 +574,42 @@ def typhos_run(
...
+
+        sys.exit(0)
 
-        if exit_after is not None and exit_after >= 0:
-            QtCore.QTimer.singleShot(exit_after * 1000.0, exit_early)
+    if exit_after is not None and exit_after >= 0:
+        QtCore.QTimer.singleShot(exit_after * 1000.0, exit_early)
 
-        return launch_suite(suite, initial_size=initial_size)
+    return launch_suite(suite, initial_size=initial_size)
```
ZLLentz, line 607: I appreciate the cleanup/dedentation you did here

typhos/cli.py
```diff
@@ -534,34 +574,42 @@ def typhos_run(
...
 
 
 def typhos_cli(args):
     """Command Line Application for Typhos."""
-    args = parser.parse_args(args)
+    args = typing.cast(TyphosArguments, parser.parse_args(args))
```
ZLLentz, line 612: There is also a built-in feature of argparse to fill a namespace instead of creating one. I'm not sure which way is better from a maintainability and type-checking standpoint.
```
args = parser.parse_args(args, TyphosArguments())
```

typhos/display.py
```diff
@@ -1405,6 +1405,9 @@ def add_device(self, device, macros=None):
         self.macros = self._build_macros_from_device(device, macros=macros)
         self.load_best_template()
 
+        if not self.windowTitle():
+            self.setWindowTitle(getattr(device, "name", ""))
```
ZLLentz, line 1409: Out of curiosity: which edge case is this fixing?

typhos/utils.py
```diff
@@ -1704,3 +1704,62 @@ def set_no_edit_style(object: QtWidgets.QLineEdit):
                 "QLineEdit { background: transparent }"
             )
         object.setReadOnly(True)
+
+
+def take_widget_screenshot(widget: QtWidgets.QWidget) -> Optional[QtGui.QImage]:
```
ZLLentz, line 1709: I like how these utils aren't actually typhos-specific and may have broader use in the future

ZLLentz (approved):
Looks great to me

I think the only thing missing is something you pointed out yourself: docs about the format strings that are accepted in the screenshot filename.

typhos/cli.py
klauer: Ah, thanks for that tip! I think your way is better than the type cast :+1:

typhos/display.py
klauer: All widgets in Qt can have window titles, as you already know, in case they're displayed as-is in the window manager.

`TyphosDeviceDisplay` or its subclasses may choose to customize the window title but they don't currently. This says "if there's no title set, at least give it the device name"

Then to answer your actual question: this title is then used during the screenshot process, with output saying "Taking a screenshot of (suite title) (device display title)":
`[2023-08-09 09:40:41] - INFO - Saving screenshot of 'Typhos Suite - EpicsMotor': 'EpicsMotor' to 'EpicsMotor.png'`

klauer:
One CI job timed out during the test suite - re-running it prior to merging

PR #567, opened 2023-08-09
FIX: patch over positioner `limits = (None, None)` issue

### LINKED ISSUE #565: Positioner widget uncaught type error in `_link_limits_by_limits_attr` (fee_m2h)
## Current Behavior
```
[2023-08-08 13:59:35] - ERROR - Unable to add 'fee_m2h' to TyphosSuite
  File "/cds/home/k/klauer/Repos/typhos/typhos/suite.py", line 734, in from_devices
    suite.show_subdisplay(device)
  File "/cds/home/k/klauer/Repos/typhos/typhos/suite.py", line 440, in show_subdisplay
    widget = self.get_subdisplay(widget)
  File "/cds/home/k/klauer/Repos/typhos/typhos/suite.py", line 421, in get_subdisplay
    subdisplay = subdisplay()
  File "/cds/home/k/klauer/Repos/typhos/typhos/display.py", line 1531, in from_device
    display.add_device(device, macros=macros)
  File "/cds/home/k/klauer/Repos/typhos/typhos/display.py", line 1416, in add_device
    self.load_best_template()
  File "/cds/home/k/klauer/Repos/typhos/typhos/display.py", line 1298, in load_best_template
    self._update_children()
  File "/cds/home/k/klauer/Repos/typhos/typhos/display.py", line 1348, in _update_children
    widget.add_device(device)
  File "/cds/home/k/klauer/Repos/typhos/typhos/panel.py", line 750, in add_device
    self._panel_layout.add_device(device)
  File "/cds/home/k/klauer/Repos/typhos/typhos/panel.py", line 857, in add_device
    self.add_sub_device(sub_device, name=dotted_name)
  File "/cds/home/k/klauer/Repos/typhos/typhos/panel.py", line 842, in add_sub_device
    container.add_device(device)
  File "/cds/home/k/klauer/Repos/typhos/typhos/display.py", line 1416, in add_device
    self.load_best_template()
  File "/cds/home/k/klauer/Repos/typhos/typhos/display.py", line 1298, in load_best_template
    self._update_children()
  File "/cds/home/k/klauer/Repos/typhos/typhos/display.py", line 1348, in _update_children
    widget.add_device(device)
  File "/cds/home/k/klauer/Repos/typhos/typhos/positioner.py", line 698, in add_device
    super().add_device(device)
  File "/cds/home/k/klauer/Repos/typhos/typhos/positioner.py", line 422, in add_device
    self._link_limits_by_limits_attr()
  File "/cds/home/k/klauer/Repos/typhos/typhos/positioner.py", line 319, in _link_limits_by_limits_attr
    if low_limit < high_limit:
TypeError: '<' not supported between instances of 'NoneType' and 'NoneType'
```

Limits can be `None`

klauer:
## Description
* `obj.limits` can be `(None, None)` in some (perhaps?) disconnected/misconfigured devices.
* These limits may not have an associated component, and may be derived directly from the control layer (i.e., EPICS CA PV low/high limits)
* This PR detects the presence of `None` and *currently* hides the limits

For discussion: on connection, it's possible those limits could go to an actual not-`None` value. In that scenario, this PR would have already hidden the limit widgets and they would not be displayed later. 

## Motivation and Context
Aims to close #565

## How Has This Been Tested?
With the "all device screenshot test" mentioned in side channels

Issue and PR text

### DIFF
--- a/typhos/positioner.py
+++ b/typhos/positioner.py
@@ -314,7 +314,11 @@ class TyphosPositionerWidget(
         except Exception:
             ...
         else:
-            if low_limit < high_limit:
+            if low_limit is None or high_limit is None:
+                # Some devices may erroneously report `None` limits.
+                # TyphosPositioner will hide the limit labels in this scenario.
+                ...
+            elif low_limit < high_limit:
                 self.ui.low_limit.setText(str(low_limit))
                 self.ui.high_limit.setText(str(high_limit))
                 return
--- a/docs/source/upcoming_release_notes/565-poslimits.rst
+++ b/docs/source/upcoming_release_notes/565-poslimits.rst
@@ -1,0 +1,23 @@
+565 poslimits
+#############
+
+API Changes
+-----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- Avoid uncaught ``TypeError`` when ``None`` is present in a positioner
+  ``.limits``.
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- klauer


klauer:
Any thoughts on what we should do here, @tangkong @ZLLentz ?

ZLLentz:
I'm doing a short investigation to understand why and in which cases this would be explicitly `(None, None)` instead of just `None` or `(0, 0)`

ZLLentz:
OK, I see it now- we hit `(None, None)` if all the PVs in a `PVPositioner` are disconnected, or if at least there was no limits arg passed and the setpoint is disconnected. In a vanilla `EpicsMotor` we raise a `DisconnectedError` instead, in our `EpicsMotorInterface` class this ends up being `(0, 0)` because we do some post-normalization.

ZLLentz:
One more important piece of context information: this is also just a fallback for when we don't have signals that directly report the low and high limit values.

typhos/positioner.py
```diff
@@ -314,7 +314,10 @@ def _link_limits_by_limits_attr(self):
         except Exception:
             ...
         else:
-            if low_limit < high_limit:
+            if low_limit is None or high_limit is None:
+                # TODO: a better fix? need to wait for connection?
+                ...
```
ZLLentz, line 319: It might be possible to have one more fallback that is subscribing to the setpoint PV's metadata callback- but other than that, there's not that much more we can do if we want to rely on an attribute instead of a signal.

klauer: I think you're right. What gave me pause mostly was the fact that we hide the limit widgets in this scenario - perhaps we just say "well, show the widgets at least so they can be set"

... In typing the above, though, I think there's no method for setting `.limits` with these widgets (in the fallback scenario you accurately described [in this comment](https://github.com/pcdshub/typhos/pull/567#issuecomment-1672144529)) - right? Or am I missing it?

ZLLentz: Right, these widgets are read-only. I think they are just the little text tags that go under the limit switch indicators.

typhos/positioner.py
```diff
@@ -314,7 +314,10 @@ def _link_limits_by_limits_attr(self):
         except Exception:
             ...
         else:
-            if low_limit < high_limit:
+            if low_limit is None or high_limit is None:
+                # TODO: a better fix? need to wait for connection?
```
klauer, line 318:
```suggestion
                # Some devices may erroneously report `None` limits.
                # TyphosPositioner will hide the limit labels in this scenario.
```

tangkong (approved):
This looks right to me, pending pre-release notes.  

I wonder if there are similar portions of typhos that have issues with unfilled/un-initialized signals/values.  I suppose we'll find them as we go

ZLLentz (approved):

PR #570, opened 2023-08-18
ENH: dynamic font resizing functionality

klauer:
## Description
* Based on tests in https://github.com/pcdshub/qtpy-dynamic-font-size-label (thanks to https://github.com/jonaias/DynamicFontSizeWidgets) this adds a monkeypatching tool for widgets which have fonts and contain text
* Using this method, the widget size in the layout then determines how big the widget will be and therefore how big the font will be
* It seems viable, though should be used sparingly as it has the potential to impact performance
* Added padding option to the original proof-of-concept

## Motivation and Context
New embedded positioner screens in #563 - @ZLLentz was aiming to have positioner readback values display in large form uniformly on all displays.

## How Has This Been Tested?
A bit interactively and with the included unit test.
It'd be a good idea to do more interactive testing, I think.
I also updated our [upstream test repository](https://github.com/pcdshub/qtpy-dynamic-font-size-label) in the meantime with the same tweaks, so it's a lot easier to test there: 

This PR.

The readback widget has this applied - and the font fills it well:
image

### DIFF
--- a/typhos/widgets.py
+++ b/typhos/widgets.py
@@ -22,7 +22,7 @@ from qtpy.QtCore import Property, QObject, QSize, Qt, Signal, Slot
 from qtpy.QtWidgets import (QAction, QDialog, QDockWidget, QPushButton,
                             QToolBar, QVBoxLayout, QWidget)
 
-from . import plugins, utils, variety
+from . import dynamic_font, plugins, utils, variety
 from .textedit import TyphosTextEdit  # noqa: F401
 from .tweakable import TyphosTweakable  # noqa: F401
 from .variety import use_for_variety_read, use_for_variety_write
@@ -392,6 +392,18 @@ class TyphosLabel(pydm.widgets.PyDMLabel):
         default = (self.displayFormat == DisplayFormat.Default)
         if new_unit.lower() in EXPONENTIAL_UNITS and default:
             self.displayFormat = DisplayFormat.Exponential
+
+    @Property(bool, "dynamicFontSize")
+    def dynamic_font_size(self) -> bool:
+        """Dynamically adjust the font size"""
+        return dynamic_font.is_patched(self)
+
+    @dynamic_font_size.setter
+    def dynamic_font_size(self, value: bool):
+        if value:
+            dynamic_font.patch_widget(self)
+        else:
+            dynamic_font.unpatch_widget(self)
 
 
 class TyphosSidebarItem(ParameterItem):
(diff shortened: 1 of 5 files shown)


klauer:
Is this worth pursuing @ZLLentz or should we aim for a different approach?

ZLLentz:
I think this is worth pursuing, not just for the readback but for the other visible text fields in the positioner widget (both the compact version and the non-compact version)

My only critique is the lack of margins which plays poorly with the blue borders (but probably looks exactly correct for non-blue-border elements)

klauer:
> My only critique is the lack of margins which plays poorly with the blue borders (but probably looks exactly correct for non-blue-border elements)

Added a padding setting for that purpose ("pad_percent", proportional to widget size) in the patching method.

I'll try to extract this from the positioner PR now - apologies for the bumpy force pushes ahead

typhos/widgets.py
```diff
@@ -393,6 +393,18 @@ def unit_changed(self, new_unit):
         if new_unit.lower() in EXPONENTIAL_UNITS and default:
             self.displayFormat = DisplayFormat.Exponential
 
+    @Property(bool, "dynamicFontSize")
```
klauer, line 396: This isn't really used and probably should be removed
I erroneously thought that `TyphosLabel` et al were in the designer, but these are really used exclusively for auto-generated device displays (where we have the ophyd signal at instantiation time, etc)

ZLLentz (approved):
I'm very :+1: on all of this
We can decide in a separate PR later how many and which widgets should get this applied to them inside the positioner widgets

tangkong (approved):
I like this a lot, my testing of various random devices this was subtle but noticeable.  

I do think this could be applied sparingly, not just for performance reasons, but also to be intentional with emphasis.  (We could pick specific important fields to allow dynamic resizing on)

Do we think there's value in a ctrl + + / ctrl + - type of feature?  Assuming it could be implemented without as bad of a performance penalty (maybe we don't have to hook paintEvent, but rather just scale font sizes up and down), that might provide users better control over sizes.

klauer:
> Do we think there's value in a ctrl + + / ctrl + - type of feature? Assuming it could be implemented without as bad of a performance penalty (maybe we don't have to hook paintEvent, but rather just scale font sizes up and down), that might provide users better control over sizes.

I think that's a good idea, though rather hard to implement properly - one widget may resize as you had hoped and another may be clipped and unreadable.

I think context-sensitive menus in PyDM, allowing fine-grained font size adjustment on a per-widget basis, could be an option. Context menus are rather underutilized there in my opinion. Imagine an "Advanced" section that allows you to increase the (fixed) size of a given widget and/or increase the font size to match its size. I think of PyDM as an expert-facing app of sorts and giving more knobs to the user (so long as they're unobtrusive) could only be a positive thing.

PR #571, opened 2023-08-18
ENH/FIX: display resizing when new templates are shown

klauer:
## Description
Automatically resize the typhos suite width when initially loaded or when a new template is loaded.

## Motivation and Context
* Horizontal scroll bars are annoying when loading up a new suite
* Aiming to alleviate that by better specifying minimum size hints and growing free-floating suites when new templates are selected

## How Has This Been Tested?
Interactively and with a small included test

This PR text

image

### DIFF
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -993,6 +993,7 @@ class TyphosDeviceDisplay(utils.TyphosBase, widgets.TyphosDesignerMixin,
 
     device_count_threshold = 0
     signal_count_threshold = 30
+    template_changed = QtCore.Signal(object)
 
     def __init__(
         self,
@@ -1287,6 +1288,16 @@ class TyphosDeviceDisplay(utils.TyphosBase, widgets.TyphosDesignerMixin,
 
         self._update_children()
         utils.reload_widget_stylesheet(self)
+        self.updateGeometry()
+        self.template_changed.emit(template)
+
+    def minimumSizeHint(self) -> QtCore.QSize:
+        if self._scroll_area is None:
+            return super().minimumSizeHint()
+        return QtCore.QSize(
+            self._scroll_area.viewportSizeHint().width(),
+            self.sizeHint().height(),
+        )
 
     @property
     def display_widget(self):
--- a/typhos/display.py
+++ b/typhos/display.py
@@ -1296,7 +1296,7 @@ class TyphosDeviceDisplay(utils.TyphosBase, widgets.TyphosDesignerMixin,
             return super().minimumSizeHint()
         return QtCore.QSize(
             self._scroll_area.viewportSizeHint().width(),
-            self.sizeHint().height(),
+            super().minimumSizeHint().height(),
         )
 
     @property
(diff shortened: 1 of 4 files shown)


typhos/display.py
```diff
@@ -1287,6 +1288,16 @@ def size_hint(*args, **kwargs):
...
         utils.reload_widget_stylesheet(self)
+        self.updateGeometry()
+        self.template_changed.emit(template)
+
+    def minimumSizeHint(self) -> QtCore.QSize:
+        if self._scroll_area is None:
+            return super().minimumSizeHint()
+        return QtCore.QSize(
+            self._scroll_area.viewportSizeHint().width(),
+            self.sizeHint().height(),
```
ZLLentz, line 1299: Does this implicitly map `minimumSizeHint`'s height to `sizeHint` in these cases? What's the effect of that (over using e.g. `super().minimumSizeHint().height()`

typhos/suite.py
```diff
@@ -462,6 +473,20 @@ def show_subdisplay(self, widget):
...
 
+        self._new_template()
+        if isinstance(widget, TyphosDeviceDisplay):
+            widget.template_changed.connect(self._new_template)
+        return widget
+
+    def _new_template(self, template: Optional[pathlib.Path] = None) -> None:
+        """Hook for when a new template is selected in a sub-display."""
+        if self.parent() is not None:
+            return
```
ZLLentz, line 484: This means that it only takes effect for standalone typhos suites, right? Rather than when it gets embedded in another app like lucid or lightpath? I think this is probably the right call, though I wonder if those apps need to handle their subdisplay widths themselves.

ZLLentz (approved):
I think this is really good and will make typhos a lot more usable

typhos/suite.py
klauer: Exactly right

I figured that we can assume what the user wants if the suite isn't embedded (`parent = human user`), but should leave it up to the parent widget otherwise.

typhos/display.py
klauer: I had found that the `minimumSizeHint` was really small in my testing for the display (like 100px or so), switched it to `sizeHint` and forgot to loop back.

`minimumSizeHint` is probably correct here. I'll switch it back and see if things are still good or if we need to re-evaluate

klauer: After switching it back, it still looks good to me. I think it didn't affect much because:
1. The dimension in question the height (where we're only resizing the width)
2. The size hint certainly is hit when laying out the device display in a scroll box, but I'm not sure about the min size hint

ZLLentz: Gotcha, that makes sense to me.

typhos/suite.py
tangkong: This was my first thought on seeing this PR 👍

tangkong (approved):
This looks good to me, and worked in my couple interactive tests.  (though the changes were slight).  

Seems like the  sizeHints passed through don't fully expand to fit sub-devices, which makes sense to me.  
Screenshot 2023-08-20 at 3 31 30 PM

PR #572, opened 2023-08-28
REF: button bar rework - template switch menu + Jira reporting button

### LINKED ISSUE #524: Disambiguate duplicate template filenames

### LINKED ISSUE #569: Typhos Jira report widget unused for over 2 years
## Expected Behavior
Someone would have reported an issue with the built-in Jira reporting widget

## Current Behavior
No one has reported an issue since the inception in #445 

## Possible Solution
Add prominent "bug" button next to the help button

klauer:
## Description
* Make template switch menu a bit more intuitive by listing classes and their corresponding displays to choose
* Add a "Jira report button" to the top level
* Remove buttons for swapping directly to different template types

## Motivation and Context
* Template selection is confusing
* Closes #569 (or tries to address it by making Jira a separate tool button)
* Closes #524 

## How Has This Been Tested?
* Interactively
* Jira reporting still works

PR text

The new toolbar:
image
`!` being "create jira report": 
image ->  image
`...` shows all templates from all categories:
image


klauer:
Alternatively, a bit more drastic of a change:
image

I'd kinda like to get rid of the buttons that were once described as "too scary to click on" in user feedback

ZLLentz:
I think the fewer-buttons look is pretty nice fwiw

tangkong:
I think this looks pretty great, much simpler.  One thought I had was maybe we could let the user know which template they're using.  I'm not sure how easy that would be (as in I don't know how I'd implement it off hand), but I'm envisioning something like this:
image

Otherwise the user might one without knowing how to get back to the template they were using

klauer:
Great suggestion @tangkong - pushed a commit with this:
image

(Styling may need some work)

typhos/display.py
klauer, line 1: Thoughts about this ordering of template types +  instead?
image

tangkong: I think it's nice! It helps make the template organization/hierarchy a little more obvious.

Leads me to wonder if people would ever want the base templates.  I'm sure there's a world where we do, but for now people probably don't care huh.

klauer: Yeah, there probably is little reason to display those - though maybe better to display more options than less just in case? The menu should look less cluttered for a non-development checkout version, but not much better admittedly.

Another thought is that this menu could also have nested submenus. I'm fond of seeing all the choices in a big list personally. That said, base/"advanced" ones still could be in an "Other screens" menu, maybe...

tangkong (approved):
I think this looks good, we could quibble about the exact format of template menu but I think it's a clear improvement in its current state

typhos/display.py
```diff
@@ -34,6 +36,15 @@ class DisplayTypes(enum.IntEnum):
...
     engineering_screen = 2
 
+    @property
+    def friendly_name(self) -> str:
+        """A user-friendly name for the display type."""
+        return {
+            self.embedded_screen: "Embedded",
+            self.detailed_screen: "Detailed",
+            self.engineering_screen: "Engineering",
+        }[self]
```
ZLLentz, line 46: It's wild to me that this works, the code reads pretty confusingly if you forget that this is an `IntEnum` (as I did, on initial review).

I don't think many other classes would be able to look themselves up in a dictionary of their own attributes.

typhos/display.py
```diff
@@ -1102,29 +1138,90 @@ def _move_display_to_layout(self, widget):
...
+
+                base_menu.addSection(f"{cls.__name__}")
+                for filename in sorted(templates, key=by_match_order):
+                    add_template(filename)
+
+        add_header("Typhos default screens")
+        for template in DEFAULT_TEMPLATES_FLATTEN:
+            add_template(template)
+
+        prefix = os.path.commonprefix(
```
ZLLentz, line 1213: Why is this widely useful utility function hiding in the os.path module? Python weird

ZLLentz (commented):
Nothing to suggest about the code- just two unnecessary observations to distract you.

I'll run with the branch and take a look at it now

typhos/display.py
klauer: I've used this pattern before with enums and kinda liked it - but if you think it's confusing it's worth redoing honestly as "clever code" is not good code

typhos/display.py
klauer: I couldn't agree more...

typhos/display.py
ZLLentz: I don't necessarily think it's worth changes- the only reason I was confused is because:

- The diff cut off at a point where I didn't see "IntEnum"
- I briefly forgot that IntEnum can be used as an int

typhos/display.py
ZLLentz: Quick note: the most up-to-date version of this looks even better with the shared prefixes cleaned up
I'm curious how this will end up looking for dev_conda

ZLLentz, line 1: Maybe unrelated to this PR and more as a consequence of #563: there's some edge cases now where you can accidentally pick a template that doesn't itself contain a button title bar, and there's no way to go back. Not sure how to handle this case.

ZLLentz (approved):
I'm :+1: on all the changes introduced here

typhos/display.py
klauer: That's a good one... maybe the suite needs to be smarter and insert a title bar if it doesn't exist in the newly-created display?

ZLLentz: Yeah, either a smart title bar or some alternate way to switch templates- right clicking?
Or, prevent us from swapping to the "dead end" templates I guess (somehow)

PR #573, opened 2023-08-28
REF/FIX: Fix Python 3.11 support and test suite-focused cleanup

klauer:
## Description
* `ophyd.Kind` `enum.Enum`, being an `IntFlag` is a bit of a special beast. Its functionality in Python 3.11 differs in enumeration of items - resulting in typhos only picking up a couple of component kinds.
* Secondarily, `multiprocessing` to run a caproto IOC ended up being problematic for Python 3.11, so I restructured the test suite a bit to run IOCs by name (simply regenerating the test cases in the parent process and in the IOC)
* Profiling code is skipped on 3.11 
* "All QMenu escape hatch" remains for only one test (all screenshots test). Unclear why this is but we get a green checkmark.
* Vendored pydm `load_ui_file` and modified it so we can always get our `Display` instance back and clean it up if desirable.
* Ignore deleted qt objects on `SignalConnection.remove_connection`
* Avoid hundreds of warnings during profiling
* Avoid creating subdisplays during "hide_subdisplays"
* Add a big pytest configuration helper to aid in finding dangling widgets
* Avoid failing screenshot taking when widgets are garbage collected at the same time

## Motivation and Context
Python 3.11 support would be nice eventually

This PR manifested itself like this:
* Python 3.11 had some obvious incompatibilities with enums, so I fixed that
* Once that was fixed, the benchmark/profiling test suite was causing major issues, so I refactored it to avoid `multiprocessing` and isolate the test IOC in its own script
* That still wasn't enough to get the test suite to pass. I saw widgets leaking from one test to another, so I added a conftest pytest hook that aimed to remove all stragglers
* Then I found that Python 3.11 benchmarking still wasn't working great, so I disabled it at moved on
* ... Then I found a bunch of stragglers and some weird things (listed above as 'avoid' or 'fix')
* And finally, there was one failing test I couldn't resolve (`test_take_top_level_widget_screenshots` sees hundreds of QMenu instances at the top level)
* And now, green checkmark

Overall, I think this is in a better spot than before. It's absolutely not perfect, though.

## How Has This Been Tested?
Locally and test suite

```python
    def dump(obj, file, protocol=None):
        '''Replacement for pickle.dump() using ForkingPickler.'''
>       ForkingPickler(file, protocol).dump(obj)
E       _pickle.PicklingError: Can't pickle <class 'ophyd.device.FlatConnect'>: attribute lookup FlatConnect on ophyd.device failed

```
image


klauer:
Well it failed spectacularly on GitHub Actions, whereas it passed locally so 🤷 
Going to mark profiling code as xfail temporarily (PR is still draft)

klauer:
OK, this is a bit of a mess and there are a ton of changes. I did this over a long period of time with (sometimes) limited attention being paid to what I was doing.

Each change should be really scrutinized!

typhos/tests/conftest.py
```diff
@@ -68,6 +72,141 @@ def noapp(monkeypatch):
...
+
+    cleanup_text = (
+        f"Not all widgets were cleaned up during {item.name}:\n"
+        + "\n".join(sorted(cleanup_descriptions))
+    )
+    failure_text = f"{item.nodeid}: {cleanup_text}"
+
+    if final_widgets:
+        if all(isinstance(widget, QtWidgets.QMenu) for widget in final_widgets):
+            logger.error("%s: Top level QMenu widgets were not cleaned up. Not failing the test suite.", item.nodeid)
```
klauer, line 195: This actually only happens in *one test*!
`test_take_top_level_widget_screenshots`
It doesn't make sense to me why this is the case.
Still, this escape hatch is what is allowing the test suite to pass for all versions of Python.

typhos/tests/conftest.py
```diff
@@ -68,6 +72,141 @@ def noapp(monkeypatch):
...
+            final_widgets.clear()
+        else:
+            logger.error(failure_text)
+
+    try:
+        assert not final_widgets, failure_text
```
ZLLentz, line 201: I like the methodology you used here to track these issues down

ZLLentz (approved):
I'm very :+1: on this
I probably left a bunch of questions as I went through this

ZLLentz:
I notice that you left some of the cleanup for another day- I think that's completely appropriate. You've cleaned up almost all of it already and the last QMenus seem to be lurking from upstream code.

typhos/tests/conftest.py
tangkong, line 201: I actually had to try this out, as it wasn't obvious to me we were asserting on both "arguments".  Python by python-ing

tangkong (approved):
The results speak for themselves, "cleanup" has never belonged more in a PR title.  

Morbid curiosity prompts me to look more closely at the test failures, but that shouldn't really hold this up, particularly since we're trying to tag in the near future

typhos/display.py
```diff
@@ -1323,10 +1329,18 @@ def _load_template(self, filename):
...
+                  else utils.load_ui_file)
 
         logger.debug('Load template using %s: %r', loader.__name__, filename)
-        return loader(str(filename), macros=self._macros)
+        try:
+            return loader(str(filename), macros=self._macros)
+        except Exception as ex:
+            display: Optional[pydm.Display] = getattr(ex, "pydm_display", None)
+            if display is not None:
+                display.setObjectName("_typhos_test_suite_ignore_")
```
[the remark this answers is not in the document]
klauer: Functionally, setting the object name for a `pydm.Display` does nothing - it's a big of a magic string for the test suite and for us when looking in with debug tools is all.

The context of this block is:
* The new template fails to load
* typhos notices that and tries to reload the prior template

In typhos master,
* A `pydm.Display` is left dangling as a top-level widget and `TyphosDeviceDisplay` never even sees a reference to it

In this PR,
* That `pydm.Display` is seen by `TyphosDeviceDisplay` and intercepted as a failed load
* We mark it for deletion and wave goodbye
* The test suite *still* picks it up because it's a top-level widget and somehow doesn't get destroyed in time for it to be happy (*see below*)
* The test suite sees the "magic" object name and says "OK this is known and expected; let's not fail the test"

Now to be honest, the marked step is fishy and it'd be nice to understand why that's happening. Is there a reference we're missing? It's worth investigating at some point.

typhos/tests/test_utils.py
```diff
@@ -276,5 +276,12 @@ def test_take_top_level_widget_screenshots(qtbot: pytestqt.qtbot.QtBot):
     widget = QWidget()
     qtbot.addWidget(widget)
     screenshots = list(utils.take_top_level_widget_screenshots(visible_only=False))
```
tangkong, line 278: A potentially dumb question.  Why is this generating any QMenus at all?  A naive (and honestly idealistic) reading of this test is that we're creating a single `QWidget`, then taking a screenshot of it.  Knowing that widgets aren't really getting cleaned up between tests, I can see why QMenus get involved, but it shouldn't generate any... right?

typhos/panel.py
```diff
@@ -711,17 +713,22 @@ def __init__(self, parent=None, init_channel=None):
         self._show_names = []
         self._omit_names = []
         # Add default Kind values
-        self._kinds = dict.fromkeys([kind.name for kind in Kind], True)
+        self._kinds = {
+            "normal": True,
+            "hinted": True,
+            "config": True,
+            "omitted": True,
+        }
```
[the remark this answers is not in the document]
klauer: `dir(Kind)` / `inspect.getmembers` / ... yeah, for something simple like this I think typing it out is sadly the best option.

Do we rely on this iterability in our other code bases, I wonder?

typhos/tests/conftest.py
```diff
@@ -68,6 +72,141 @@ def noapp(monkeypatch):
...
+        return
+
+    for widget in widgets:
+        widget = widget()
+        if widget is None:
+            continue
+
+        try:
+            widget.isVisible()
+            widget.windowTitle()
```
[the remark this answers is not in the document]
klauer: I think calling one may be sufficient and my reasoning here is a bit hand-wavy/anecdotal:
* It seems like simple boolean attributes may not trigger the "is the object still alive check; if not raise RuntimeError to alert the user"
* So I added `windowTitle()` as a second one, since it requires an actual string to Python string conversion under the hood

It'd be good to dig into the details one day, but for today I think double method call is good enough for me...

typhos/tests/conftest.py
klauer: You mean `assert x, y` vs `assert (x, y)`? That's a common one that trips people up for sure
Stashing the complicated message (`y`) here kinda necessitated this form

typhos/tests/test_utils.py
klauer: I have repeated this ~dumb~ question to myself many times over the course of this PR.

I have no answers 😦 
That said, there's one thing I didn't try: what if we load up gammaray when this test fails and we poke around with it there? It'd be worthwhile I think

klauer:
🤞

typhos/tests/conftest.py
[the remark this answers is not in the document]
tangkong: I'm totally good with being doubly-sure.  Just curious if there was some portion of the widget that doesn't get caught by `widget.isVisible()` or something

typhos/tests/conftest.py
tangkong: I usually end up just doing `assert x`, `assert y`.

typhos/tests/conftest.py
```diff
@@ -68,6 +72,141 @@ def noapp(monkeypatch):
...
+) -> List[QtWidgets.QWidget]:
+    res = []
+    for obj in objs:
+        obj = obj()
+        if obj is not None:
+            res.append(obj)
+    return res
+
+
+@pytest.hookimpl(hookwrapper=True, trylast=True)
```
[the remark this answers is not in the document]
tangkong: Does a test failing at teardown make all the subsequent tests fail? 
* test passes (fail at teardown) -> test fails -> test passes
or 
* test passes (fail at teardown) -> test fails -> test fails....

typhos/tests/conftest.py
klauer: Oh, but that's actually not what this is doing - check for yourself:
In the above `x` is the condition and `y` is the message to report to the user
`assert False, "hello"` - doesn't mean `all([False, "hello"])` but rather: if not `False`, show message "hello"

Or according to the Python docs, which are overly complicated here: `if not expression1: raise AssertionError(expression2)` (ref: https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement
)

tangkong: Oh, I thought I did. And I'm not sure why I thought my previous understanding made any sense 💀 ignore me 😞

klauer: No worries - glad that cleared it up though

typhos/plugins/core.py
```diff
@@ -94,6 +95,10 @@ def __init__(self, channel, address, protocol=None, parent=None):
         # Add listener
         self.add_listener(channel)
 
+    def __dtor__(self) -> None:
```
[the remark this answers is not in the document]
klauer: Sadly no... this is a weird (undocumented?) internal part of sip which, right before the C++ object gets deleted, this Python method gets called: https://github.com/search?q=__dtor__+language%3APython&type=code

typhos/tests/conftest.py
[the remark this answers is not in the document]
klauer: Fixtures failing at teardown cause only the subsequent test to fail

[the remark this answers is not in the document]
klauer: For the record:
* The reason why this PR doesn't include a *fixture* to do the same check is due to the above: fixture teardown exceptions don't affect the current test but the (single) subsequent test
* The pytest hook as implemented of this PR functions as we would hope: the assertion in this hook applies to the active test (and does not interfere with the subsequent test)

klauer:
2 green checkmarks in a row indicate it's probably good enough as-is.
Before this gets any larger, if you guys are good with it - I'd say let's merge @tangkong @ZLLentz 

We may revisit and adjust how strict this is in the future based on how many intermittent failures we get (due to gc/teardown slowness). Increasing the wait time may also suffice...

typhos/panel.py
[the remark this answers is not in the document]
ZLLentz: neither of those standard functions actually gives you the results which is wild to me
```
>>> 'omitted' in dir(Kind)
False
>>> Kind.omitted
<Kind.omitted: 0>
```
(I forgot to install ipython in my py311 test branch)

ZLLentz:
I'll click merge now to make sure it happens

PR #582, opened 2023-09-08
ENH: positioner timeout scaling

ZLLentz:
## Description

Increase the motion timeout times proportionally for longer moves, to decrease the chances of a false positive.

## Motivation and Context

There were some complaints from MFX that their motors were timing out early.

Prior to this PR, the motion timeouts were:

- Minimum 5 second (settle_time)
- otherwise, slightly greater than the expected travel time (distance/speed + 2*acceleration time)

In some cases for longer moves, the expected time to get to the goal would be wrong by greater than 5 seconds.
This makes it so that these cases are less likely to frivolously error without compromising faster timeout feedback for the shorter moves. I toyed with increasing the base settle time from 5 to, say, 10, but that doesn't help these long moves enough and it delays the timeout message for motors that are expected to arrive "immediately".

I also considered including the motor's settle_time parameter but I wanted to get out a small PR without any unnecessary baggage and without needing more in-depth testing.

## How Has This Been Tested?

The tests all still pass and the code here is fairly simple

Added pre-release notes

### DIFF
--- a/typhos/positioner.py
+++ b/typhos/positioner.py
@@ -203,7 +203,7 @@ class TyphosPositionerWidget(
         thread.status_finished.connect(self._status_finished)
         thread.start()
 
-    def _get_timeout(self, set_position, settle_time):
+    def _get_timeout(self, set_position: float, settle_time: float, rescale: float = 1) -> float | None:
         """Use positioner's configuration to select a timeout."""
         pos_sig = getattr(self.device, self._readback_attr, None)
         vel_sig = getattr(self.device, self._velocity_attr, None)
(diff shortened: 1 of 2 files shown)


typhos/positioner.py
```diff
@@ -222,7 +222,7 @@ def _get_timeout(self, set_position, settle_time):
         else:
             acc_time = acc_sig.get()
         # This time is always greater than the kinematic calc
-        return abs(delta/speed) + 2 * abs(acc_time) + abs(settle_time)
+        return rescale * (abs(delta/speed) + 2 * abs(acc_time)) + abs(settle_time)
```
tangkong, line 225: If we're not going to include settle time maybe it's worth explaining ourself a bit in the docstring here... though I guess it's not really a public method.  I feel like it's a 50/50 shot someone will think the `rescale` applies to `settle_time` as well

tangkong (approved):

typhos/positioner.py
ZLLentz: I'll write docstrings, they were glaringly missing prior to this edit and I was hoping I wouldn't need to but I should just do it

My thought was that you wouldn't want your two input args to interfere with each other

PR #586, opened 2023-09-12
FIX: make setpoint work for original positioner widget, adjust sizing

### LINKED ISSUE #585: State positioner devices in the non-row (classic?) form do not get visible combo boxes.
## Expected Behavior

There should be a combo box selector.

## Current Behavior

Instead, there's a blank space. Incidentally this mucks up the widget layout.

image

I originally thought this was a templating or expert button issue but I no longer believe this is the case.

## Context

Blocker for deploying recent row updates (backwards compatibility issue).

## Your Environment

typhos master, pcdsdevices master

ZLLentz:
## Description

Fixes an issue where the setpoint widgets for the detailed views were coming up with zero width.

The root cause of the issue was that the sizing of the user_setpoint placeholder widget was set to the defaults, so the `setMinimumWidth` call was setting the minimum width of the combobox to 0.

The reason why these various widgets refuse to expand here despite having their size policy set to "Expanding" eludes me, so I've set them to fixed with.

This also incidentally combines some previously diverging code paths that fixes a sizing issue where the detailed view had a vestigial/unused pydm line edit taking up space.

## Motivation and Context

closes #585 

## How Has This Been Tested?

Interactively only

I need to make a release notes docs entry

Master:
image

This PR:
image

### DIFF
--- a/typhos/positioner.py
+++ b/typhos/positioner.py
@@ -476,10 +476,6 @@ class TyphosPositionerWidget(
             self.ui.set_value.addItems(setpoint_signal.enum_strs)
             # Activated signal triggers only when the user selects an option
             self.ui.set_value.activated.connect(self.set)
-            self.ui.set_value.setSizePolicy(
-                QtWidgets.QSizePolicy.Expanding,
-                QtWidgets.QSizePolicy.Fixed,
-            )
             self.ui.set_value.setMinimumContentsLength(20)
             self.ui.tweak_widget.setVisible(False)
         else:
@@ -487,14 +483,23 @@ class TyphosPositionerWidget(
             self.ui.set_value.setAlignment(QtCore.Qt.AlignCenter)
             self.ui.set_value.returnPressed.connect(self.set)
 
+        self.ui.set_value.setSizePolicy(
+            QtWidgets.QSizePolicy.Fixed,
+            QtWidgets.QSizePolicy.Fixed,
+        )
+        self.ui.set_value.setMinimumWidth(
+            self.ui.user_setpoint.minimumWidth()
+        )
         self.ui.set_value.setMaximumWidth(
             self.ui.user_setpoint.maximumWidth()
         )
-
         self.ui.setpoint_layout.addWidget(
             self.ui.set_value,
             alignment=QtCore.Qt.AlignHCenter,
         )
+        self.ui.set_value.setObjectName('set_value')
+        # Because set_value is used instead
+        self.ui.user_setpoint.setVisible(False)
 
     @property
     def device(self):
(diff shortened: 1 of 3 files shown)


typhos/ui/widgets/positioner.ui
```diff
@@ -319,17 +319,23 @@ Screen</string>
...
            </sizepolicy>
           </property>
           <property name="minimumSize">
            <size>
-            <width>0</width>
+            <width>201</width>
             <height>25</height>
            </size>
           </property>
+          <property name="maximumSize">
```
tangkong, line 333: With dynamic sizing I wonder if setting maximum sizes will ever come back to bite us.  Admittedly I don't remember enough of the dynamic sizing details to answer this myself...

typhos/positioner.py
```diff
@@ -476,25 +476,30 @@ def _define_setpoint_widget(self):
...
+        )
         self.ui.set_value.setMaximumWidth(
             self.ui.user_setpoint.maximumWidth()
         )
-
         self.ui.setpoint_layout.addWidget(
             self.ui.set_value,
             alignment=QtCore.Qt.AlignHCenter,
         )
+        self.ui.set_value.setObjectName('set_value')
```
tangkong, line 500: I know Ken was doing this for testing, but should we be doing this more uniformly in our UI's?  What does setting the object name here do?

tangkong (approved):
This works for me, good catch.  Had a few questions but no problems with the contents at large

typhos/ui/widgets/positioner.ui
ZLLentz: It turns out that these input sub-widgets don't resize at all anyway, even with the size policy set to "expanding". I don't have any idea why this is the case. Before I went to fixed mode, the widget wasn't even taking up all of the available space.

In addition:
For the row widget: we want the readback to expand instead of the setpoint
For the square widget: we don't want anything to resize, e.g. we don't want to see the whole middle section balloon out and we don't want the square to stretch up and down.

typhos/positioner.py
ZLLentz: There are two things this does:
1. The widget now has a name when inspected via gammaray. This is the only reason I did this here, technically I could remove it. I suspect this could also be helpful for tracebacks/other debug like Ken was doing for testing.
2. There are some qt functions for collecting child widgets based on their name: https://doc.qt.io/qt-6/qobject.html#findChild, and they use this qt object name rather than the python attribute name

klauer: Additionally, it can be used as a method of differentiating widgets in the stylesheet

PR #587, opened 2023-09-13
MNT: add TyphosDisplaySwitcher to positioner_row.ui

### LINKED ISSUE #580: Switching to templates without title bars
> Maybe unrelated to this PR and more as a consequence of #563: there's some edge cases now where you can accidentally pick a template that doesn't itself contain a button title bar, and there's no way to go back. Not sure how to handle this case.

_Originally posted by @ZLLentz in https://github.com/pcdshub/typhos/pull/572#discussion_r1318895338_

tangkong:
## Description
Adds TyphosDisplaySwitcher to row positioner widget.

This only required changes to the .ui file, to my complete amazement.  👏 

## Motivation and Context
closes #580 in this case, but not the general one.

## How Has This Been Tested?
Clicking through all the other templates available to me, and I never got stuck

This PR

image

### DIFF
--- a/typhos/positioner.py
+++ b/typhos/positioner.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import inspect
 import logging
 import os.path
 import threading
@@ -9,6 +10,8 @@ from typing import Optional, Union
 import ophyd
 from pydm.widgets.channel import PyDMChannel
 from qtpy import QtCore, QtWidgets, uic
+
+from typhos.display import TyphosDisplaySwitcher
 
 from . import dynamic_font, utils, widgets
 from .alarm import AlarmLevel, KindLevel, _KindLevel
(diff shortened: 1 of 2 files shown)


tangkong:
pip py3.9 failed once with the standard flood of QMenus, but passed on rerun... :/

ZLLentz:
This is a good call
Since we added the title/notes row, there's definitely room for this now

ZLLentz (approved):
There is some oddity here with the (?) button- it doesn't seem to functional when these are subdevices. I'd expect to see e.g. the motor class docstring or the ophyd doc field. Otherwise everything here works as expected.

ZLLentz:
Actually- it seems like only for the positioner row templates, the ? button is completely non-functional. There must be some code reason for this.

ZLLentz:
Ah, so I think the code for the help button is defined only in the full title bar widget, not in this button. So we'd need to do a bit of a refactor to make the help button work as intended.

tangkong:
Ah you're totally right.  The help frame is created and added to TyphosDisplayTitle, which sets up the toggle functionality and webviewer.  

In the interest of keeping this template compact, maybe we just show the docstring via tooltip and disable clicking for this template.

tangkong:
Something like this? 
image

ZLLentz (approved):

PR #595, opened 2023-11-14
MAINT: switch typhos cli defaults to embedded/vertical by popular demand

### LINKED ISSUE #532: Feedback: Switch default view to 'embedded'
## Expected Behavior
* Feedback indicates that the default screens show too much information
* Suggestion was that 'embedded' be the default

However, if we work on the existing templates to be more compact or space-efficient may result in a different desired default.

## Current Behavior
Engineering screens are the default

## Related to
#533

ZLLentz:
## Description

When launching a typhos suite via the CLI, set the default look/feel to the embedded/vertical mode.

## Motivation and Context

In recent months we've revamped the most-used embedded templates (motors) to be row widgets, which give a much better screen in general if stacked vertically.

In the past prior to these changes, there was already clamoring for embedded as the default for screen space saving (closes #532)

## How Has This Been Tested?

Interactively only

Motors open up in neat row-by-row line widgets
Larger devices (somehow?) open up as they always have (TODO investigate)

Only here

## Pre-merge checklist

### DIFF
--- a/docs/source/upcoming_release_notes/595-maint_cli_defaults.rst
+++ b/docs/source/upcoming_release_notes/595-maint_cli_defaults.rst
@@ -1,0 +1,25 @@
+595 maint_cli_defaults
+######################
+
+API Breaks
+----------
+- Default typhos suites to "embedded" displays arranged "vertically" instead of
+  "detailed" displays arranged "horizontally" to more naturally match how typhos is
+  used at present.
+
+Features
+--------
+- N/A
+
+
+Bugfixes
+--------
+- N/A
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- zllentz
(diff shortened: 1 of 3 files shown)


ZLLentz:
This had no objections multiple times, so I'm going to finish the PR now:

- docs
- some investigation into the ci failures

ZLLentz:
I'll spend a bit of time evaluating the wide test suite failures if they happen again

ZLLentz:
One of the failing tests on widget cleanup was a case where we were using a stylesheet to set the widget template- something that is in no way expected or standard behavior. I think this was causing an infinite widget creation loop:

- Set widget stylesheet
- Force template to a non-existent template using stylesheet
- As part of the template load process, refresh the stylesheet

I've switched this test out for one that simply turns label text to red via stylesheet, which is more in line with expected user behavior.

tangkong (approved):
LGTM.   Shouting into the void is frustrating as always.

PR #602, opened 2024-04-12
DOC: compile and clarify release notes for v3.1.1

ZLLentz:
## Description

Just the limit switch color/contrast change


ZLLentz:
3.10 is busted but I'm going to punt on this because we aren't really supporting 3.10 yet locally

PR #607, opened 2024-05-31
CI/TST: Fix CI

ZLLentz:
## Description

This PR resolves the following CI issues:
- The CI fails for Python 3.10+ if QT_QPA_PLATFORM=offscreen with a segmentation fault.
- There was an additional failure related to pyqtgraph changes that resulted in the per-test widget cleanup assert failing for the singular test that creates a pyqtgraph widget. This affected only the conda builds...

The PR fixes these largely by adding a `no_gc` pytest mark that skips the `gc.collect()` call that causes some of the tests to have a segmentation fault and a `no_cleanup_check` mark that skips the post-test cleanup assert.

In short: skip these in the cases where they don't work.

The other changes here are:

- Ignore .vscode config
- Skip benchmark tests by default, because they are long-running and tend to leak into other parts of the test suite.
- Make debug log more detailed
- Slow down the repeated `gc.collect()` calls.
- Call `start_simulation` in `RandomSignal` to get rid of deprecation warning
- xfail the profiler test because it (rightfully) fails on py312 but I will need to investigate and fix it separately
- Stop using deprecated/removed unittest mock API call (called_with -> assert_called_with)

## Motivation and Context

Need to be able to rely on the results from the CI test suite for proper development flow
Needed to make sure the issues weren't "real" (affecting users)
Need to move on and work on other things

## How Has This Been Tested?

Reproduced the issue offline, understood when it happens, adjusted tests

Here only

## Pre-merge checklist

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -2,3 +2,4 @@
 filterwarnings = default
                  once::DeprecationWarning
 addopts = --benchmark-skip
+markers = no_gc: mark a test as unstable with the per-test gc
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,3 +1,4 @@
 [pytest]
-filterwarnings= default
-                once::DeprecationWarning
+filterwarnings = default
+                 once::DeprecationWarning
+addopts = --benchmark-skip
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,3 +3,4 @@ filterwarnings = default
                  once::DeprecationWarning
 addopts = --benchmark-skip
 markers = no_gc: mark a test as unstable with the per-test gc
+          no_cleanup_check: mark a test to skip the widget cleanup assert
--- a/typhos/tests/conftest.py
+++ b/typhos/tests/conftest.py
@@ -204,11 +204,14 @@ def pytest_runtest_call(item: pytest.Item):
         else:
             logger.error(failure_text)
 
-    for widget in final_widgets:
-        try:
-            widget.deleteLater()
-        except RuntimeError:
-            ...
+    try:
+        assert not final_widgets, failure_text
+    finally:
+        for widget in final_widgets:
+            try:
+                widget.deleteLater()
+            except RuntimeError:
+                ...
 
 
 @pytest.fixture(scope='session')
--- a/typhos/tests/conftest.py
+++ b/typhos/tests/conftest.py
@@ -203,7 +203,7 @@ def pytest_runtest_call(item: pytest.Item):
         "test_dialog_button_instances_smoke"
     ]
     try:
-        if item.name not in skip_test_names:
+        if all(skip_name not in item.name for skip_name in skip_test_names):
             assert not final_widgets, failure_text
     finally:
         for widget in final_widgets:
--- a/typhos/tests/conftest.py
+++ b/typhos/tests/conftest.py
@@ -183,8 +183,6 @@ def pytest_runtest_call(item: pytest.Item):
                 "\n".join((desc, "\n    -> ".join([""] + ref_desc)))
             )
         referrers.clear()
-    # Clear reference to iteration variable
-    widget = None
 
     cleanup_text = (
         f"Not all widgets were cleaned up during {item.name}:\n"
@@ -213,16 +211,6 @@ def pytest_runtest_call(item: pytest.Item):
                 widget.deleteLater()
             except RuntimeError:
                 ...
-    if final_widgets:
-        # Remove reference to widgets
-        final_widgets = None
-        app = QtWidgets.QApplication.instance()
-        # Try to prod the garbage collector
-        app.processEvents()
-        gc.collect()
-        app.processEvents()
-    # One more for the road to enter next test with clean event queue
-    app.processEvents()
 
 
 @pytest.fixture(scope='session')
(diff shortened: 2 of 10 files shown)


ZLLentz:
Next week:
- Self-review
- re-evaluate the assert that I disabled
- open to reviewers
- go back to other typhos tickets with fewer distractions

ZLLentz:
> Is the garbage collection you're turning off the manual garbage collection we placed into the typhos test suite right? Normal pytest / python garbage collection remains?

Yes, this is just the manual `gc.collect()` call that I'm disabling here

> Is there any rhyme or reason to avoiding garbage collection?

I noticed that any time two tests with positioner widgets were run in the same test suite execution, the second one would segfault on the `gc.collect()` call. I haven't thought very hard about why this is the case. This is true no matter what you do with the positioner widget, so long as the uic load step is run.

I decided that, in general, it was better to collect more garbage to avoid cross-test contamination, but since it's failing here I want to skip it and move on. Perhaps with more time we could figure out _why_ the test segfaults on `gc.collect()` in more detail.

typhos/tests/conftest.py
```diff
@@ -197,14 +204,11 @@ def pytest_runtest_call(item: pytest.Item):
         else:
             logger.error(failure_text)
 
-    try:
-        assert not final_widgets, failure_text
```
ZLLentz, line 201: This is the main thing I want to re-evaluate before asking for review. I'll fix up the desc before requesting a review too.

ZLLentz: I removed this edit because I thought it was a somewhat important check and it passed offline in all permutations.
Then, it failed again here because pyqtgraph isn't being cleaned up in a timely manner (or at all?) in the online CI only.

ZLLentz: I'm going to add another mark, apply it to the failing function, and then move on.

ZLLentz:
test failure is `api.MAMBA_DOWNLOAD_FAILFAST`
marking as ready for review, asking for a review, and then re-running the failure

ZLLentz:
Well, it didn't give me the option to only re-run the failure, so we're rerunning everything now!

ZLLentz:
Ok great! Hopefully once reviewed and merged and I can finalize some of the other PRs for review (since the test suite will be useful again)

pytest.ini
```diff
@@ -1,3 +1,6 @@
 [pytest]
-filterwarnings= default
-                once::DeprecationWarning
+filterwarnings = default
+                 once::DeprecationWarning
+addopts = --benchmark-skip
```
tangkong, line 4: Was this also a necessary fix to prevent segfaults?

typhos/tests/conftest.py
```diff
@@ -129,8 +129,15 @@ def pytest_runtest_call(item: pytest.Item):
...
-        gc.collect()
+        if not list(item.iter_markers(name="no_gc")):
+            # Some tests segfault on gc.collect when QT_QPA_PLATFORM=offscreen (!?)
+            # Mark these with pytest.mark.no_gc to skip this call
+            # Note that some other tests definitely need this gc.collect() call!
+            # And some tests function just fine no matter whether we call this or not!
+            gc.collect()
         app.processEvents()
+        # Throttle while loop to avoid weird gc things
+        time.sleep(0.1)
```
tangkong, line 140: This worries me, given inevitable hardware differences on the CI runners.

pytest.ini
ZLLentz: I don't think this was necessary to prevent segfaults, but I was super worried by seeing log messages related to the benchmark components bleeding into the other tests. Note that the benchmarking doesn't actually even run on py311, py312...

typhos/tests/conftest.py
ZLLentz: I think the size of the sleep timer here didn't matter so much, the important part was to give background threads some time to process in cases where `app.processEvents()` is short rather than repeatedly calling `gc.collect()` in a no-wait loop.

pytest.ini
ZLLentz: I think I'd prefer to wait to bring back the benchmarking until we have a more structured way to track the results over time.

tangkong (dismissed):
I think the results speak for themselves here.  The only remaining thoughts i have are:
* we should squash the commits before we merge
* We should record somewhere that using the positioner / motor fixtures require this no-gc mark.  Since I'm sure to forget that offscreen-mode is necessary to observe these errors

ZLLentz:
> We should record somewhere that using the positioner / motor fixtures require this no-gc mark. Since I'm sure to forget that offscreen-mode is necessary to observe these errors

I'm going to look at the repo and find the best spots to put this info into the development docs

Agreed on the squash, I had some interesting commit iterations here

ZLLentz:
I decided to write some prose about the test suite, including this info and some other important things.

ZLLentz:
ok I'm actually done now and not going to add another tiny commit
if the docs build comes out rendering normally and the text seems ok I think this is done

ZLLentz:
Seems to render OK
I might have angered the CI gods by causing so many test suite runs though
image

ZLLentz:
I think this is ready to be squashed now (unless I have another typo/inaccuracy/missing info to fix in the docs)

tangkong (approved):
LGTM.  On to the actual thing you wanted to do.

PR #611, opened 2024-06-06
ENH: positioner row widget sizing usability fixes and tweaks

ZLLentz:
## Description

Rework the design, sizing, and font scaling of the positioner row widget to address concerns about readability in RIX for devices with large position numbers, and to address issues with poorly used space in state devices.

#### dynamic_font.py:
- Allow dynamic font scaling to be applied to comboboxes
- Allow dynamic font resizing widgets to optionally define a minimum and maximum font size.
- Let font size change by less than the delta if it is to snap to the minimum or maximum, making sizing more consistent at the edges.
- Minor caching optimizations related to dynamic font scaling (there could be more improvements here) 
- setText-based widgets with dynamic font scaling now also adjust their font size when their text contents change
- Remove sizing overrides for dynamic patched widgets because they are no longer needed now that we're not using paint events and they get in the way of implementing proper sizing elsewhere
- Set the font size prior to the first resize in case the widget is never resized.

#### panel.py
- Fix all known sizing issues related to recursive composite signal panels

#### positioner.py
- Remove `reload_widget_stylesheet` call from `clear_error` to avoid an issue where clicking `clear_error` would undo all dynamic font scaling and set it back to the default size.
- Include the possibility for the limit indicators to completely disappear to allow other widgets to use their space
- Remove references to fixed sizing to allow for the possibility of widget scaling
- Remove unused layout modifications now that the layouts are different
- Apply dynamic font sizing to every single widget with text in the positioner row widget
- Add the low/high limit position settings to the expandable panel below the row widget
- Allow the setpoint widget to expand into the space left by the tweak widget for the combobox case
- Rework/refactor the expandable signal panel to be a pop-out window (the new `RowDetails` widget) on top of the screen instead of inside of the main layout, because adding a huge widget to the layout causes serious issues with resizing.
- Remove references to static "Error" text that is removed in the row widget ui file

#### test_dynamic_font.py
- Make sure combobox font patching works

#### PositionerBase.embedded.ui
- Set minimum size, set size policy to be expanding so that the row is allowed to expand

#### positioner.ui
- Remove all instances of fixed font-size that interfere with dynamic text resizing
- Remove all alarm-sensitive borders because I had to disable a widget style reload earlier

#### positioner_row.ui
- Completely redesign this to be aesthetically scalable in all directions
- Small note: made the clear_error button smaller to save vertical space
- Small note: switched squares to circles because the square widgets resize terribly

## Motivation and Context

- Certain device screens, such as sp1k1 and the imagers, look kind of wonky due to sizing.
- https://jira.slac.stanford.edu/browse/ECS-5339

## How Has This Been Tested?

- Locally
- Added a little bit of testing
- In person test at RIX

Need to revise the pre-release notes one more time

## Screenshots:
Before small:
image

Before wide:
image

After small:
image

After wide:
image

After default size with default size/location expansion pop-outs:
image

## Pre-merge checklist

### DIFF
--- a/typhos/dynamic_font.py
+++ b/typhos/dynamic_font.py
@@ -3,9 +3,12 @@ Dynamic font size helper utilities:
 
 Dynamically set widget font size based on its current size.
 """
+import logging
 
 from qtpy import QtCore, QtGui, QtWidgets
 from qtpy.QtCore import QRectF, Qt
+
+logger = logging.getLogger(__name__)
 
 
 def get_widget_maximum_font_size(
@@ -119,7 +122,7 @@ def patch_widget(
         maximum font size. Content margin settings determine the content
         rectangle, and this padding is applied as a percentage on top of that.
     """
-    def resizeEvent(event: QtGui.QResizeEvent) -> None:
+    def set_font_size() -> None:
         font = widget.font()
         font_size = get_widget_maximum_font_size(
             widget, widget.text(),
@@ -129,6 +132,9 @@ def patch_widget(
         if abs(font.pointSizeF() - font_size) > 0.1:
             font.setPointSizeF(font_size)
             widget.setFont(font)
+
+    def resizeEvent(event: QtGui.QResizeEvent) -> None:
+        set_font_size()
         return orig_resize_event(event)
 
     def minimumSizeHint() -> QtCore.QSize:
@@ -152,6 +158,7 @@ def patch_widget(
     widget.resizeEvent = resizeEvent
     widget.sizeHint = sizeHint
     widget.minimumSizeHint = minimumSizeHint
+    set_font_size()
 
 
 def unpatch_widget(widget: QtWidgets.QWidget) -> None:
--- a/typhos/dynamic_font.py
+++ b/typhos/dynamic_font.py
@@ -5,6 +5,7 @@ Dynamically set widget font size based on its current size.
 """
 from __future__ import annotations
 
+import functools
 import logging
 
 from qtpy import QtGui, QtWidgets
@@ -132,11 +133,13 @@ def patch_widget(
     """
     def set_font_size() -> None:
         font = widget.font()
-        font_size = get_widget_maximum_font_size(
-            widget, widget.text(),
-            pad_width=widget.width() * pad_percent,
-            pad_height=widget.height() * pad_percent,
+        font_size = get_max_font_size_cached(
+            widget.text(),
+            widget.width(),
+            widget.height(),
         )
+        # 0.1 = avoid meaningless resizes
+        # 0.00001 = snap to min/max
         delta = 0.1
         if max_size is not None and font_size > max_size:
             font_size = max_size
@@ -147,6 +150,17 @@ def patch_widget(
         if abs(font.pointSizeF() - font_size) > delta:
             font.setPointSizeF(font_size)
             widget.setFont(font)
+
+    # Cache and reuse results per widget
+    # Low effort and not exhaustive, but reduces the usage of the big loop
+    @functools.lru_cache
+    def get_max_font_size_cached(text: str, width: int, height: int) -> float:
+        return get_widget_maximum_font_size(
+            widget,
+            text,
+            pad_width=width * pad_percent,
+            pad_height=height * pad_percent,
+        )
 
     def resizeEvent(event: QtGui.QResizeEvent) -> None:
         set_font_size()
--- a/typhos/dynamic_font.py
+++ b/typhos/dynamic_font.py
@@ -213,7 +213,7 @@ def patch_text_widget(
 
     resizeEvent._patched_methods_ = (
         widget.resizeEvent,
-        widget.setText
+        widget.setText,
     )
     widget.resizeEvent = resizeEvent
     widget.setText = setText
@@ -238,8 +238,8 @@ def patch_combo_widget(
         font_sizes = [
             get_max_font_size_cached(
                 text,
-                widget.height(),
                 widget.width(),
+                widget.height(),
             )
             for text in combo_options
         ]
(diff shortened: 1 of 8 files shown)


ZLLentz:
I'm going to finish up the documentation todo stuff and then I'm going to adjust how I implemented the dynamic sizing. I've realized some basic tricks for making these sorts of composite widgets cleanly rescaleable in all directions while maintaining proportions and while only needing specification in the ui file (rather than my current code here which does a lot of extra stuff in the py files that it shouldn't need to if the ui is set up correctly). In essence: the only sizing rule that we'll need to keep track of is the dynamic font resizing, which I plan to attach to all of the text widgets instead of just some of them.

ZLLentz:
I like where this is at now

I didn't quite get to implementing the splitters idea, because the way the basic and composite signal panels are implemented (as a QWidget that is also itself a QGridLayout) makes this somewhat annoying to implement. There will need to be a sizable refactor to achieve this. (Now: add rows to a grid layout that is actually a typhos panel, later: add rows to a splitter area? And maintain the grid sizing? And only have splitters between subdevices, not between signals?)

The splitters idea will become an issue for implementing later since I think it would be helpful in a lot of cases.

I am going to clean up the description for documentation purposes but I don't plan on modifying the code again until next PR unless there are code quality issues or bugs. I think after this PR I'll switch to going through the most recent typhos issues.

ZLLentz:
This is now reviewable so I'm going to mark as ready for review. I still will revise the pre-release notes one more time.

ZLLentz:
I also want to make a companion PR on pcdsdevices, I'll try not to forget. This works by itself but there's a couple more things to tweak.

tangkong:
wowee this is a biggun

A few notes from interactive testing:
* Largely everything rendered fine on my end.  I'm not entirely sure how to mock high-dpi displays, maybe I can petition for a nice big monitor 🤔 
* Some strange clipping in non-positioner widgets.  On master this happens bug can be fixed by making the window taller, but with the changes in this PR I can't seem to reveal the rest of the readouts.
  * image
* (Unrelated to this PR) Generally these windows rendered too tall for my monitor, with no ability to scroll to the bottom.  On mac at least the only resize slider is at the bottom right, making it impossible to resize.  Even when I can resize it, the minimum widget size often doesn't let me.  I wonder if it's a scroll area angle, where we limit the size of the initial window to the monitor size
* I like the popouts.  I do wonder if we'll run into a problem of these windows getting left behind when the user clicks on the main window, bringing it into focus.  If you open one of these then close the parent typhos widget, it does stay alive (and produces a segfault).  One way to fix this is by putting this widget in a QMenuAction, which we could set to close when it loses focus.  Otherwise I suspect properly parenting this widget might be the solution
* (immensely minor nitpick) It might just be me, but I feel like a thin horizontal line between the positioner widgets would be helpful.  As is the top and bottom of the widget-group doesn't really have a clear line, so it can be ambiguous whether the "clear error" and "status" buttons might belong to the group either above or below it.

I'll give the code a read through now

typhos/dynamic_font.py
```diff
@@ -118,26 +126,153 @@ def patch_widget(
...
+
+def patch_text_widget(
+    widget: QtWidgets.QLabel | QtWidgets.QLineEdit,
+    *,
+    pad_percent: float = 0.0,
+    max_size: float | None = None,
+    min_size: float | None = None,
+):
+    """
+    Specific patching for widgets with text() and setText() methods
```
tangkong, line 165: This docstring could be a bit more helpful.  I think it'd be nice to get a quick summary of what exactly we're patching: the `setText` and `resizeEvent` methods.  

It seems like this is our entry point to patching other behavior in the future as well, so keeping tabs of exactly what's going on here will be helpful

typhos/positioner.py
```diff
@@ -1009,32 +1024,20 @@ def _create_signal_panel(self) -> Optional[TyphosSignalPanel]:
...
             return None
 
-        panel = TyphosSignalPanel()
-        panel.omitNames = self.get_names_to_omit()
-        panel.sortBy = SignalOrder.byName
-        panel.add_device(self.device)
-
-        self.ui.layout().addWidget(panel)
-        return panel
+        return RowDetails(row=self)
```
tangkong, line 1027:
```suggestion
        return RowDetails(row=self, parent=self, flags=QtCore.Qt.Window)
```

If you pass the flags kwarg through to QWidget, the little window will close when the parent closes but still show up as a separate window

```python
class RowDetails(QtWidgets.QWidget):
    """
    Container class for floating window with positioner row's basic config info.
    """
    row: TyphosPositionerRowWidget
    resize_timer: QtCore.QTimer

    def __init__(self, row: TyphosPositionerRowWidget, parent: QtWidgets.QWidget | None = None, **kwargs):
        super().__init__(parent=parent, **kwargs)
        self.row = row
```

tangkong (commented):
Hard to argue with the results here. I pointed out a few things I saw

typhos/positioner.py
ZLLentz: Thank you, I didn't know this at all

ZLLentz:
One more dumb bug I need to resolve: the text snaps back to minimum size before and after a move

ZLLentz:
Another dumb "only while moving real things" bug: the combobox text gets repeatedly bigger and bigger and bigger each time you click on it...
image

ZLLentz:
Text getting small is because pydm is manually resetting the style on certain metadata events.
It seems like the only way around this is to also set the stylesheet to have the font specifier in it, so I'll try that.

ZLLentz:
Setting the stylesheet fixed both the big big font and the tiny tiny font issues

ZLLentz:
The only remaining fix I'm going to do here is to address:

> Some strange clipping in non-positioner widgets

After that, I'll round up the remaining items into tickets:

- thin lines between positioner rows
- splitters between devices

ZLLentz:
side-by-side of old and new which shows off the sizing updates and shows how much is being clipped with the bug
image

ZLLentz:
Left: fixed clipping in default view
Right: expanding to +config view shows more and dynamically resizes
Changing the views to remove items also shrinks the widgets to remove the whitespace
image

ZLLentz:
There's only one more situation where the sizing is wrong- when the signal panel has no entries in it at all. Maybe this becomes a ticket to fix later, too.

ZLLentz:
Remaining issue is that subdevices of subdevices don't collapse properly when vanished... my first try at that one caused more issues so this might be where I leave it for now.

tangkong (approved):
I think this looks good on the whole.  We have some followups, but this is a clear improvement

PR #616, opened 2024-06-25
FIX: metadata handling for enums

### LINKED ISSUE #605: Initial value is set before enum choices in a PyDMEnumComboBox
## Expected Behavior

During initialization, a new ``pcdsdevices.plugins.core.SignalConnection`` object should set the enum_choices (really all metadata) first, then set the signal's value so that it can select from the available choices. 

## Current Behavior

When the ``PyDMEnumComboBox`` first loads, there is an error message from PyDM:

```Can not change value to 'nA'. Not an option in PyDMComboBox```

even though when I look at the widget, "nA" is one of the options. This happens regardless of what the enum_choices actually are.

I think this is because the [``SignalConnection.add_listener`` method](https://github.com/pcdshub/typhos/blob/ad6ffafb2788af6d48ae997f3c5abb1430af6443/typhos/plugins/core.py#L270)  first calls ``self.send_new_value(signal_val)`` then ``self.send_new_meta(**signal_meta)``.

After the widget is shown, If I change the signal's value outside of PyDM (e.g. though channel access), then the widget updates properly.

## Possible Solution

Reversing the order of ``self.send_new_value(…)`` and ``self.send_new_meta(…)`` fixes this problem at my beamline.

I'm happy to prepare a PR that fixes this.

1. Create a PyDM display with a PyDMEnumComboBox somewhere.
2. Set the channel of the combo box to point to an Ophyd ``EpicsSignal`` obj with enum choices
3. Load the PyDM display

## Context

I have several windows that use this pattern. Right now, we get incorrect values shown in the PyDMEnumComboBox when it first loads. This makes it difficult to use, since we first have to wait for the PV to change some other way (or force it to change somehow). It'd be nice if it showed the correct value from the beginning. 

## Your Environment

Here are the list of installed packages I'm using: https://gist.github.com/canismarko/4f681511b295451179b421a14ba0566e

Running in a conda environment on a RHEL9 computer.

### LINKED ISSUE #608: SignalConnection mis-casts the put value from a PyDMEnumComboBox if ``is_string=True``
This applies when a a PyDMEnumComboBox is set up to connect to an Ophyd signal with enum_strs, and the signal has ``is_string`` set to True.

## Expected Behavior

Changing the combo box should send the selected value to the ophyd signal, either as the string for that enum choice, or as an int for the enum index.

## Current Behavior

When changing the value, the ``SignalConnection`` tries to cast the received value (the index of the selected combobox item) to the dtype of the signal, in this case a string because of ``is_string=True``. In most cases, the ophyd signal interprets the string as the index of the enum choice, which is correct. However, if the index is also one of the possible strings in enum_choices, then the ophyd signal incorrectly interprets the put value as an enum choice.

For example, let the ``enum_strs`` for a signal be ``('1', '2', '5')``. The PyDMEnumComboBox then shows these 3 as options. If I select '5' (``enum_strs[2]``). [``SignalConnection.cast()``](https://github.com/pcdshub/typhos/blob/ad6ffafb2788af6d48ae997f3c5abb1430af6443/typhos/plugins/core.py#L125) then receives the value ``2``, and attempts to cast it to the signals type, which is a str, so converts it to ``"2"``. This value then gets passed to the signal's ``put("2")`` method. Since ``"2"`` is a valid choice, the ophyd signal then sets its own value to ``"2"`` (``enum_strs[1]``), which is incorrect.

I think this is a Typhos issue and not an Ophyd issue, because passing an index as a string is ambiguous and we should really pass indexes as an int, in which case it is properly interpreted as the index of an enum choice. It would be unreasonable to expect Ophyd to resolve this ambiguity properly in all cases.

## Possible Solution

One solution is to add a check to [``SignalConnection.cast()``](https://github.com/pcdshub/typhos/blob/ad6ffafb2788af6d48ae997f3c5abb1430af6443/typhos/plugins/core.py#L125) to handle this case. Perhaps:

```python

    enum_strs = self.signal.describe()[self.signal.name].get('enum_strs', ())
    try:
        value = enum_strs[value]
    except (TypeError, IndexError):
        pass
```

this would need to be done before the actual type-casting happens on line 149.

This solution adds some overhead because we have to try and get the enum_str even if the signal has no ``enum_strs``. I suspect this extra step is way faster that the actual ``put()`` call to the Ophyd signal, but if it's not then we could run an additional check to ensure ``len(enum_strs) > 0``, filtering out those majority of cases that don't involve enum_strs.

1. Create an ophyd Signal with ``enum_strs=('1', '2', '5'), is_string=True``.
2. Create a PyDMEnumComboBox with this signal as its channel.
3. Change the combobox to select the last entry ("5").

## Context

The EPICS support for an SRS-570 preamp has a PV for the sensitivity value (e.g. "5") and a separate one for the sensitivity unit ("uA/V"). The allowed values for the sensitivity value are ``("1", "2", "5", "10", "20", "50", "100", "200", "500")`` to match the selectable options of the pre-amp. Creating a PyDMEnumCombox for this signal triggers this bug if "2", "5" or "50" are selected because these correspond to indexes ``1``, ``2``, and ``5`` and will instead set the signal to "1", "2", and "5" respectively.

## Your Environment

Mamba environment: https://gist.github.com/canismarko/08a3c73535151c8a19b76d13b8be710d

Running on RHEL 9.

ZLLentz:
## Description

Fix three issues related to enum/metadata handling in the `SignalConnection` part of the `SignalPlugin` and implement simplified unit tests for these cases.

- Switch the order of the metadata/value setup for new listeners to ensure widgets have the correct contextual information. This is exactly the fix proposed by @canismarko in #605.
- Fix an issue where string=True enum EpicsSignals would reliably break on casting in cases where the enum strings were numerically the same as other indices in the enum_strs tuple. This builds upon the fix proposed by @canismarko in #608
- Fix a related (unreported) issue where the previous point could also break for string=False EpicsSignals.

## Motivation and Context

closes #605 
closes #608 

## How Has This Been Tested?

Unit tests were added. I originally tried to make a larger unit test where I instantiated real/fake EpicsSignal instances and connected them to the combobox widgets as reported in the issues, but I couldn't get this to work properly in a unit testing context. Rather than dump additional dev hours into this, I came up with simpler tests, possibly these could be better.

I will need to test with a real EPICS PV too.

I need to write some release notes

## Pre-merge checklist

### DIFF
--- a/typhos/plugins/core.py
+++ b/typhos/plugins/core.py
@@ -83,6 +83,8 @@ class SignalConnection(PyDMConnection):
         self._connection_open = True
         self.signal_type = None
         self.is_float = False
+        self.enum_strs = ()
+
         # Collect our signal
         self.signal = self.find_signal(address)
         # Subscribe to updates from Ophyd
@@ -143,7 +145,24 @@ class SignalConnection(PyDMConnection):
                          dtype, self.signal.name, self.signal_type)
 
         logger.debug("Casting %r to %r", value, self.signal_type)
-        if self.signal_type is np.ndarray:
+        if self.enum_strs:
+            # signal_type is either int or str
+            # use enums to cast type
+            if self.signal_type is int:
+                # Get the index
+                try:
+                    value = self.enum_strs.index(value)
+                except (TypeError, ValueError, AttributeError):
+                    value = int(value)
+            elif self.signal_type is str:
+                # Get the enum string
+                try:
+                    value = self.enum_strs[value]
+                except (TypeError, ValueError):
+                    value = str(value)
+            else:
+                raise TypeError(f"Invalid combination: {self.enum_strs=} with {self.signal_type=}")
+        elif self.signal_type is np.ndarray:
             value = np.asarray(value)
         else:
             value = self.signal_type(value)
@@ -227,6 +246,7 @@ class SignalConnection(PyDMConnection):
             self.unit_signal.emit(units)
         if enum_strs is not None:
             self.enum_strings_signal.emit(enum_strs)
+            self.enum_strs = enum_strs
 
         # Special handling for severity
         if severity is None:
@@ -266,9 +286,9 @@ class SignalConnection(PyDMConnection):
         else:
             self.is_float = False
 
-        # Report new value
-        self.send_new_value(signal_val)
+        # Report new meta for context, then value
         self.send_new_meta(**signal_meta)
+        self.send_new_value(signal_val)
         # If the channel is used for writing to PVs, hook it up to the
         # 'put' methods.
         if channel.value_signal is not None:
--- a/typhos/plugins/core.py
+++ b/typhos/plugins/core.py
@@ -80,10 +80,10 @@ class SignalConnection(PyDMConnection):
     def __init__(self, channel, address, protocol=None, parent=None):
         # Create base connection
         super().__init__(channel, address, protocol=protocol, parent=parent)
-        self._connection_open = True
-        self.signal_type = None
-        self.is_float = False
-        self.enum_strs = ()
+        self._connection_open: bool = True
+        self.signal_type: type | None = None
+        self.is_float: bool = False
+        self.enum_strs: tuple[str, ...] = ()
 
         # Collect our signal
         self.signal = self.find_signal(address)
@@ -161,7 +161,9 @@ class SignalConnection(PyDMConnection):
                 except (TypeError, ValueError):
                     value = str(value)
             else:
-                raise TypeError(f"Invalid combination: {self.enum_strs=} with {self.signal_type=}")
+                raise TypeError(
+                    f"Invalid combination: enum_strs={self.enum_strs} with signal_type={self.signal_type}"
+                )
         elif self.signal_type is np.ndarray:
             value = np.asarray(value)
         else:
--- a/docs/source/upcoming_release_notes/616-fix_enum_handling.rst
+++ b/docs/source/upcoming_release_notes/616-fix_enum_handling.rst
@@ -1,0 +1,23 @@
+616 fix_enum_handling
+#################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- Fix various issues with enum handling in the SignalPlugin.
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- canismarko
+- zllentz
--- a/typhos/tests/plugins/test_core.py
+++ b/typhos/tests/plugins/test_core.py
@@ -5,8 +5,9 @@ import pydm.utilities
 import pytest
 from ophyd import Component as Cpt
 from ophyd import Device, Signal
+from ophyd.sim import EnumSignal
 from pydm import PyDMApplication
-from pydm.widgets import PyDMLineEdit
+from pydm.widgets import PyDMChannel, PyDMLineEdit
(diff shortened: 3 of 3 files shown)


ZLLentz:
Interactively, applying this fixed the mix-up between the enum values for putting for a ("1", "2", "5") enum signal but it actually increased the number of error messages I got at startup...

ZLLentz:
Scratch that- the extra spam errors come from the signal not being fully connected when the widget comes online. Waiting for connection resolves the spam. This isn't something to overlook but it's unrelated to the problem the PR is fixing.

typhos/plugins/core.py
```diff
@@ -83,6 +83,8 @@ def __init__(self, channel, address, protocol=None, parent=None):
         self._connection_open = True
         self.signal_type = None
         self.is_float = False
+        self.enum_strs = ()
```
tangkong, line 86: I got confused for a bit about what type this actually ends up as, it's not always a tuple right?  Maybe a type hint here would help, since it's buried in `send_new_meta`?

typhos/plugins/core.py
```diff
@@ -143,7 +145,24 @@ def cast(self, value):
...
+                except (TypeError, ValueError, AttributeError):
+                    value = int(value)
+            elif self.signal_type is str:
+                # Get the enum string
+                try:
+                    value = self.enum_strs[value]
+                except (TypeError, ValueError):
+                    value = str(value)
+            else:
+                raise TypeError(f"Invalid combination: {self.enum_strs=} with {self.signal_type=}")
```
tangkong, line 164: TIL you could do this with f-strings.  This ends up also printing `self`, might not be the most helpful.  But this is truly a nitpick

tangkong (dismissed):
This looks right to me.  All I have to offer is documentation type nitpicks.  

Did we ever run into this on our end?  Or did we just never set `string` at all?

typhos/plugins/core.py
ZLLentz: Let me check, I think it's always a tuple but it might also be a list

typhos/plugins/core.py
ZLLentz: Oh shoot I didn't realize this would point self... I'll make some sort of adjustment

ZLLentz:
Locally at LCLS we hadn't hit this yet for a couple reasons:

- Most of our screens that use EPICS go through the PyDM EPICS plugins instead of through the Typhos signal plugins
- We typically hadn't been setting string=True

typhos/plugins/core.py
ZLLentz: It looks like `enum_strs` is always casted to a tuple when we receive them from the control layer before filling the `metadata` dictionary, and are again casted to a tuple in `describe`. I'll annotate this as a tuple.

ZLLentz:
Interestingly, as I was testing this, I found that there was also an equivalent issue for `string=False` signals, but whether this caused an issue depended on the precise implementation of the widget that we're connecting to. In the case of PyDM's combobox the values are sent through as integers, so the int-native signal was unaffected and the string-native signal was broken. A widget which sent the values through as strings would have broken int-native signals in the same way.

typhos/plugins/core.py
ZLLentz: Simple rework to put the variable names in myself:
```
TypeError: Invalid combination: enum_strs=('zero', 'one', 'two') with signal_type=<class 'float'>
```

tangkong (approved):
LGTM

PR #620, opened 2024-10-08
ENH/MNT/TST: make auto scrollable functional again

### LINKED ISSUE #599: Scroll bar defaults are no longer sensible
## Expected Behavior

I should get a scrollbar when I need one, e.g. if the windows are large compared to my screen size

## Current Behavior

- "scrollable" is a cli option that defaults to "auto" which means "false if embedded, true otherwise".
- Embeded is now the default style, so screens open by default with no scrollbar, even if they are tall tree displays
- Suites with dozens of embedded screens open vertically now with no scrollbar

## Possible Solution

- Stop doing scroll like this?
- Scrollable defaults to true for suites with >1 device?
- Use the scroll if needed features that are native to qt

ZLLentz:
## Description

- Make scrollable=auto choose to include the scollbar or not based on the chosen template's display type. Previously, it was the selected type, even if that loaded a differently typed template. This means that "embedded" displays that open up large trees will now automatically be given scrollbars.
- Add some helpers to accomplish the above
- Add some tests to verify the above
- Switch display/suite defaults to match the cli defaults

## Motivation and Context

closes #599 
as part of https://jira.slac.stanford.edu/browse/ECS-6333

## How Has This Been Tested?

I added some tests and ran some devices interactively.

Only here so far, and in the pre-release notes.
I was going to rewrite the cli help text but the original text more closely matches the new behavior than the old behavior.
```
parser.add_argument(
    '--scrollable',
    default='auto',
    help=(
        'Whether or not to include the scrollbar. '
        'Valid options are "auto", "true", "false", '
        'and any unique shortenings of those options. '
        'Selecting "auto" will include a scrollbar for '
        'non-embedded layouts.'
    ),
)
```

## Pre-merge checklist

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -20,6 +20,7 @@ requirements:
   build:
   - python >=3.9
   - pip
+  - setuptools
   - setuptools_scm
   host:
   - python >=3.9
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -22,9 +22,6 @@ requirements:
   - pip
   - setuptools
   - setuptools_scm
-  host:
-  - python >=3.9
-  - pip
   run:
   - python >=3.9
   - coloredlogs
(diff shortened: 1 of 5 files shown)


ZLLentz:
I'll need to tweak the conda recipe slightly thanks to the setuptools thing

ZLLentz:
Score another point for "making it more similar to the other thing"

tangkong (approved):
Looks good to me.  No objections here.

PR #621, opened 2024-10-09
FIX: formalize hotfix for resizing empty comboboxes

ZLLentz:
## Description

This is a hotfix we've been using for a while at LCLS.
It helps us avoid raising an exception the dynamic font rescaling function when it is applied to a combobox with no entries.

The normal logic flow here is to calculate the maximum allowable font size for each possible combobox entry, then take the minimum of these results so that any of the entries would fit in the combobox.

When there are no entries, we can't take the `min` of an empty list, leading to an exception.
Instead, the code will now exit since there's no reason to resize anything if there is no text.

This is an incomplete fix, the complete version would also include a dynamic resize when the combobox options are changed.
The reason this fix was never completed is that, after the IOC that caused this problem was fixed, we no longer had a test case to develop further code against. This test case also never yielded us valid enum strings to work with, so the only part that could be verified is the part that stopped the screen from completely crashing on the exception.

This was deemed "good enough" because it stops the screen from crashing at least. I think it's also good enough to be included in the default branch.

## Motivation and Context

This prevents these edge case malformed enum cases from crashing screens (instead: empty combobox).

## How Has This Been Tested?

Interactively and in production at LCLS for many months.

Here only

## Pre-merge checklist
- [ ] ~New/changed functions and methods are covered in the test suite where possible~

### DIFF
--- a/docs/source/upcoming_release_notes/621-fix_combobox_resize_crash.rst
+++ b/docs/source/upcoming_release_notes/621-fix_combobox_resize_crash.rst
@@ -1,0 +1,23 @@
+621 fix_combobox_resize_crash
+#############################
+
+API Breaks
+----------
+- N/A
+
+Features
+--------
+- N/A
+
+Bugfixes
+--------
+- Fix an issue where dynamic font scaling applied to a combobox
+  with no entries would raise an exception and close typhos.
+
+Maintenance
+-----------
+- N/A
+
+Contributors
+------------
+- zllentz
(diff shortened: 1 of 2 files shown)


tangkong (approved):
Yea seems fine to me.  Another instance of patching external problems in python, but that's our fate I suppose

ZLLentz:
I agree that this is in response to "something else" behaving poorly, but I think someone could accidentally apply this font resizing functionality to an empty combobox manually. In that hypothetical case, I think the non-crashing behavior is good!

PR #627, opened 2025-02-06
Add long names

aberges-SLAC:
## Description
Made changes to `SignalPanel` class  in `display` module to support `long_names` for device signals. This approach should be _pretty_ close to compatible with https://github.com/bluesky/ophyd/pull/1187, so if & when that comes through we should only need minimal changes.

## Motivation and Context
Have you ever had a screen with 30+ signals and maybe multiple subcomponent devices with identical signal names?  Maybe you wanted alternative signal labels for cleaner GUI that end users can play with?

Figured some of the work I did in https://github.com/pcdshub/pcdsdevices/pull/1320 could be generalized in `typhos` and potentially trim the fat out of my screen scripts. 

Long names lets you use an optional string to replace the `label_text` in a `signal row` object added to your grid of signals. When you use a `long_name`, it will add the signal's `dotted_name` to the tooltip so you can still hover and get hutch python context.

## How Has This Been Tested?
I didn't have any unit tests to run, per se, but I did test this on a variety of devices for screens we've got out in prod (and some I'm cooking in a different PR). 

I was able to open the entire CRIXS MODS screen without everything crashing and burning, so that's gotta amount to something, right?

https://github.com/user-attachments/assets/845e1877-e2c5-4409-9680-1f46b9946226

This PR! And in the code itself

## Pre-merge checklist
- [ ] Code works interactively
- [ ] Code contains descriptive docstrings, including context and API
- [ ] New/changed functions and methods are covered in the test suite where possible
- [ ] Code has been checked for threading issues (no blocking tasks in GUI thread)
- [ ] Test suite passes locally
- [ ] Test suite passes on GitHub Actions
- [ ] Ran ``docs/pre-release-notes.sh`` and created a pre-release documentation page
- [ ] Pre-release docs include context, functional descriptions, and contributors as appropriate

### DIFF
--- a/typhos/panel.py
+++ b/typhos/panel.py
@@ -250,7 +250,7 @@ class SignalPanel(QtWidgets.QGridLayout):
         tooltip: str
             Doc string to add to signal
         long_name: str, optional
-            Long form (human readable) name to use for the signal row label text.    
+            Long form (human readable) name to use for the signal row label text.
         """
         if long_name:
             label_text = long_name
@@ -268,7 +268,7 @@ class SignalPanel(QtWidgets.QGridLayout):
 
     def _get_long_name(self, device, attr, dotted_name) -> str:
         """
-        Check the signal for its long_name, if it exists. 
+        Check the signal for its long_name, if it exists.
         Until Ophyd makes it a standard signal, need to manually check
         the device and its components for the name.
 
@@ -294,7 +294,6 @@ class SignalPanel(QtWidgets.QGridLayout):
             if hasattr(getattr(device, dotted_name), 'long_name'):
                 long_name = getattr(device, dotted_name).long_name
         return long_name
-
 
     def add_signal(self, signal, name=None, long_name=None, *, tooltip=None):
         """
@@ -328,7 +327,7 @@ class SignalPanel(QtWidgets.QGridLayout):
             return
 
         logger.debug("Adding signal %s (%s)", signal.name, name)
-        
+
         label = self._create_row_label(attr=name, dotted_name=name, long_name=long_name, tooltip=tooltip)
         loading = utils.TyphosLoading(
             timeout_message='Connection timed out.'
--- a/typhos/panel.py
+++ b/typhos/panel.py
@@ -266,7 +266,7 @@ class SignalPanel(QtWidgets.QGridLayout):
             label.setToolTip(_tooltip)
         return label
 
-    def _get_long_name(self, device, attr, dotted_name) -> str:
+    def _get_long_name(self, device, attr, dotted_name):
         """
         Check the signal for its long_name, if it exists.
         Until Ophyd makes it a standard signal, need to manually check
--- a/typhos/panel.py
+++ b/typhos/panel.py
@@ -235,16 +235,38 @@ class SignalPanel(QtWidgets.QGridLayout):
                for _, sig_info in signal_pairs):
             self.loading_complete.emit([name for name, _ in signal_pairs])
 
-    def _create_row_label(self, attr, dotted_name, tooltip):
-        """Create a row label (i.e., the one used to display the name)."""
-        label_text = self.label_text_from_attribute(attr, dotted_name)
+    def _create_row_label(self, attr, dotted_name, tooltip, long_name=None):
+        """
+        Create a row label (i.e., the one used to display the name).
+        If an alternative (human-readable) long name is defined, use that instead
+        and add the dotted_name to the tooltip for hutch python ease of use.
+
+        Parameters
+        -----------
+        attr: any
+            Name of the signal
+        dotted_name: any
+            Full dotted name of the signal
+        tooltip: str
+            Doc string to add to signal
+        long_name: str, optional
+            Long form (human readable) name to use for the signal row label text.    
+        """
+        if long_name:
+            label_text = long_name
+        else:
+            label_text = self.label_text_from_attribute(attr, dotted_name)
         label = SignalPanelRowLabel(label_text)
         label.setObjectName(dotted_name)
         if tooltip is not None:
-            label.setToolTip(tooltip)
+            if long_name:
+                _tooltip = dotted_name + '<br>' + round(1.75*len(dotted_name))*'-' + '<br>' + tooltip
+            else:
+                _tooltip = tooltip
+            label.setToolTip(_tooltip)
         return label
 
-    def add_signal(self, signal, name=None, *, tooltip=None):
+    def add_signal(self, signal, name=None, long_name=None, *, tooltip=None):
         """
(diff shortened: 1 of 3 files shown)


typhos/panel.py
```diff
@@ -332,8 +354,18 @@ def _add_component(self, device, attr, dotted_name, component):
...
 
+        # Hacky workaround until Ophyd.Component.long_name PR comes through
+        long_name = None
+        try:
+            if hasattr(getattr(device, attr), 'long_name'):
+                long_name = getattr(device, attr).long_name
+        except AttributeError:
+            # Then maybe we have a nested component and can't touch the signal
+            if hasattr(getattr(device, dotted_name), 'long_name'):
+                long_name = getattr(device, dotted_name).long_name
```
ZLLentz, line 365: If this bit is a hacky workaround, and we needed it twice, it should be pulled out into its own function. After pulling out into its own function this is a reasonable thing to write a short unit test for to make sure we've covered all the cases we expect.

I'm wondering if there's some way to revise this to make it more readable but I don't have any immediate suggestions. Some of the hasattr calls probably aren't needed I guess.

typhos/panel.py
ZLLentz, line 1: Maybe missing is the pre-release notes docs step, same as how we've done it in pcdsdevices with [pre-release-notes.sh](https://github.com/pcdshub/typhos/blob/master/docs/pre-release-notes.sh)

ZLLentz (commented):
I like the way this is implemented. I think the docs inclusion is required before merge and the other suggestion is optional.

typhos/panel.py
aberges-SLAC: Ah derp forgot to commit that, let me fix that after above changes

typhos/panel.py
aberges-SLAC: Agreed, it's probably worth some:

```
def _get_long_name(self, device, attr, dotted_name) -> str:
"""
Check if the signal has implemented a long_name and return it. Will check both the device and the component signal for the long_name if it exists.
"""
    long_name = None
    try:
            if hasattr(getattr(device, attr), 'long_name'):
                long_name = getattr(device, attr).long_name
        except AttributeError:
            # Then maybe we have a nested component and can't touch the signal
            if hasattr(getattr(device, dotted_name), 'long_name'):
                long_name = getattr(device, dotted_name).long_name
    return long_name
```

I think I ended up needing both `hasattr` due to it getting squirrelly when you build a screen with multiple component devices where the long name is actually in `device.component.attr.long_name` instead of `device.attr.long_name` — and the fact that `long_name` is not in the Ophyd object by default yet. Once they implement it we can probably just replace it with some `try` block around `if getattr(device, attr).long_name` instead.

typhos/panel.py
```diff
@@ -662,8 +694,17 @@ def _maybe_add_signal(self, device, attr, dotted_name, component):
...
                                dotted_name, device.name, ex, exc_info=True)
                 return
-
-            return self.add_signal(signal, name=attr, tooltip=component.doc)
+            # Hacky workaround until Ophyd.Component.long_name PR comes through
+            long_name = None
+            try:
+                if hasattr(getattr(device, attr), 'long_name'):
+                    long_name = getattr(device, attr).long_name             
+            except AttributeError as ex:
```
ZLLentz, line 702: Small note: the pre-commit CI job is flagging ex as an unused variable and a could instances of trailing whitespace

typhos/panel.py
```diff
@@ -266,6 +266,36 @@ def _create_row_label(self, attr, dotted_name, tooltip, long_name=None):
...
+            device: (any)
+                The Ophyd.Device with component signals
+            attr: (str)
+                The str name of the signal attribute
+            dotted_name: (str)
+                The full dotted name to the signal
+
+        Returns
+        --------
+            str or None
```
ZLLentz, line 286: Super petty spacing nitpick to match the other docstrings
```suggestion
        device: (any)
            The Ophyd.Device with component signals
        attr: (str)
            The str name of the signal attribute
        dotted_name: (str)
            The full dotted name to the signal

        Returns
        --------
        str or None
```

typhos/panel.py
```diff
@@ -266,6 +266,36 @@ def _create_row_label(self, attr, dotted_name, tooltip, long_name=None):
...
+        """
+        long_name = None
+        try:
+            if hasattr(getattr(device, attr), 'long_name'):
+                long_name = getattr(device, attr).long_name
+        except AttributeError:
+            # Then maybe we have a nested component and can't touch the signal
+            if hasattr(getattr(device, dotted_name), 'long_name'):
+                long_name = getattr(device, dotted_name).long_name
+        return long_name
```
ZLLentz, line 296: Super petty optional don't-allocate-the-variable edit
```suggestion
        try:
            if hasattr(getattr(device, attr), 'long_name'):
                return getattr(device, attr).long_name
        except AttributeError:
            # Then maybe we have a nested component and can't touch the signal
            if hasattr(getattr(device, dotted_name), 'long_name'):
                return getattr(device, dotted_name).long_name
```

typhos/panel.py
```diff
@@ -266,6 +266,36 @@ def _create_row_label(self, attr, dotted_name, tooltip, long_name=None):
             label.setToolTip(_tooltip)
         return label
 
+    def _get_long_name(self, device, attr, dotted_name) -> str:
```
ZLLentz, line 269: More pettiness: I normally wouldn't ask you to annotate, but this one is half annotated. Can you either finish the annotations or remove them?
device: typing.Any
attr: str
dotted_name: str
-> str | None

ZLLentz (approved):
I've only got nitpicks left
If you're happy and finished we're good to go

I still like the implementations here and the result

typhos/panel.py
aberges-SLAC: I'll remove the annotations to be more consistent with the rest of the script, although it probably wouldn't hurt to add the annotations to all of the funcs one day

aberges-SLAC:
> I've only got nitpicks left If you're happy and finished we're good to go
> 
> I still like the implementations here and the result

Should pass the checks now, gave it the ole flake8 rundown. Passing `black` over it changed quite a few things so I'm gonna pass on that to save us a lot of frantic debugging

ZLLentz (approved):
Looks good to me, let's roll
Remaining CI is an unrelated weird python 3.11 race condition that I don't need to think about today

typhos/panel.py
tangkong: I don't think this ever got committed 👀

ZLLentz: whoops

aberges-SLAC: I don't even remember getting this message on the commit o.O

PR #635, opened 2025-11-25
BUG/MNT: add value opt to pyqtgraph.parametertree.Parameter and its subclasses

tangkong:
## Description
- Ensures `value` is provided at `__init__` for all subclasses of `pyqtgraph.parametertree.Parameter`

## Motivation and Context
Gotta love those breaking updates!  In fairness this was as deprecation warning we didn't clock, and they've just deprecated it a month early.

We subclass `Parameter` once, so there we set a `value=None` as default.  In other places where we use built-in types, we initialize `value=None` as well.

## How Has This Been Tested?
By getting the test suite to pass. 

I've also checked in the test suite that the values we collect are the same, by inspecting the `items` dictionary from [here](https://github.com/pcdshub/typhos/blob/master/typhos/suite.py#L875).  both before and after this patch, the form of the dictionary is the same (only the object ids change)

This PR, and pre-release notes

## Pre-merge checklist

### DIFF
--- a/typhos/suite.py
+++ b/typhos/suite.py
@@ -53,6 +53,8 @@ class SidebarParameter(parametertree.Parameter):
     sigEmbed = QtCore.Signal(object)
 
     def __init__(self, devices=None, embeddable=None, **opts):
+        if "value" not in opts:
+            opts["value"] = None
         super().__init__(**opts)
         self.embeddable = embeddable
         self.devices = list(devices) if devices else []
@@ -266,7 +268,7 @@ class TyphosSuite(TyphosBase):
 
         self._tree = parametertree.ParameterTree(parent=self, showHeader=False)
         self._tree.setAlternatingRowColors(False)
-        self._save_action = ptypes.ActionParameter(name='Save Suite')
+        self._save_action = ptypes.ActionParameter(name='Save Suite', value=None)
         self._tree.addParameters(self._save_action)
         self._save_action.sigActivated.connect(self.save)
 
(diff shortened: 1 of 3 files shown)


tangkong:
marking as draft until I figure out why the conda tests specifically hang

tangkong:
TLDR: this is ready to merge, please ignore the conda failure

I've tracked down the source of the conda errors, and they are unrelated to the changes made here.  Fixing it will require adjustments to ci-helpers.  

To summarize, coverage has been shown to cause timeouts in test suites, for more recent coverage versions and python versions.  The reason this doesn't appear on the pip tests is because we only run conda with coverage.  Local testing with coverage managed to reproduce the hangs, and my test matrix is as follows:

- (coverage=7.6.1, pytest-cov=5.0.0, py3.9): success   (pcds-5.9.1)
- (coverage=7.6.1, pytest-cov=5.0.0, py3.12): fail 
- (coverage=7.6.1, pytest-cov=6.3.0, py3.12): fail 
- (coverage=7.11.0, pytest-cov=7.0.0, py3.12): fail 
- (coverage=7.12.0, pytest-cov=7.0.0, py3.12): fail 

Basically, I wasn't able to find a py3.12 configuration that worked.  I plan to remove the coverage option from the conda test and separate it into a separate optional job

typhos/suite.py
ZLLentz, line 1: Changes seem reasonable but there's a lot of lingering dev artifacts in the test suite

tangkong: whoops sorry those should be cleaned up now.  Maybe with the ci-helpers update this will actually run

ZLLentz:
It looks like the pip test is exploding now

ZLLentz (approved):

tangkong:
The coverage test will fail due to the coverage issue I referenced in https://github.com/pcdshub/pcds-ci-helpers/pull/167 , but the pip tests on their own pass.  Unfortunately I wasn't able to actually fix the coverage-enabled tests, but I'm not sure it's possible without digging into coverage.

